PR #3, opened 2018-09-06
make Trace class public

qinfchen:
This is needed to delegate all remoting tracing calls to `tracing-java`. Since [Tracer.getAndClearTrace()](https://github.com/palantir/tracing-java/blob/develop/tracing/src/main/java/com/palantir/tracing/Tracer.java#L246-L251) is public and returns Tracer type, it makes sense for Tracer to be public


iamdanfox (approved):

iamdanfox:
I think this is an acceptable tradeoff - we haven't seen nearly as much abuse of tracing libraries compared to http-remoting, and I think the overall outcome of seamlessly switching http-remoting to tracing-java is very valuable.

PR #9, opened 2018-09-11
compatibility check against tracer remoting

qinfchen:
tracing-java now check remoting tracing version to ensure all tracings are done in a single thread, see more details https://github.com/palantir/http-remoting/pull/799. If remoting version is lower than 3.43.0, an exception will be thrown. 

I am not sure what is the best way of testing incompatible case, may be some integration test specifically for that. Any suggestions?

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -42,9 +42,13 @@ import org.slf4j.MDC;
  * This class is thread-safe.
  */
 public final class Tracer {
+
     private static final Logger log = LoggerFactory.getLogger(Tracer.class);
     private static final OrderableSlsVersion MIN_COMPAT_REMOTING_TRACER_VER = OrderableSlsVersion.valueOf("3.43.0");
     private static final Optional<OrderableSlsVersion> REMOTING_TRACER_VER = getRemotingTracerVersion();
+
+    // system property used to disable compatibility check against remoting tracer
+    private static final String REMOTING_TRACER_COMPAT_CHECK_SYSTEM_PROPERTY = "tracing.remoting.compat.check.enabled";
 
     private Tracer() {
         checkForRemotingTracerCompatibility(REMOTING_TRACER_VER);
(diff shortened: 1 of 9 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -274,4 +304,23 @@ static void setTrace(Trace trace) {
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
     }
 
+    protected static void checkForRemotingTracerCompatibility(Optional<OrderableSlsVersion> remotingTracerVer) {
+        if (remotingTracerVer.isPresent()
+                && VersionComparator.INSTANCE.compare(remotingTracerVer.get(), MIN_COMPAT_REMOTING_TRACER_VER) < 0) {
+            throw new IllegalStateException(String.format(
```
dansanduleac: can we use `com.palantir.safe-logging:preconditions` and `Checks.checkState` ?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -77,6 +81,8 @@ private static Trace createTrace(Optional<Boolean> isObservable, String traceId)
      * #setSampler configured sampler} returns true.
      */
     public static void initTrace(Optional<Boolean> isObservable, String traceId) {
+        checkForRemotingTracerCompatibility(REMOTING_TRACER_VER);
```
dansanduleac: Why don't we do this statically, on class load?
Might as well fail immediately if we have an incompatible classpath

dansanduleac (changes requested):

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -274,4 +304,23 @@ static void setTrace(Trace trace) {
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
     }
 
+    protected static void checkForRemotingTracerCompatibility(Optional<OrderableSlsVersion> remotingTracerVer) {
```
iamdanfox, line 288: I think it's pretty important that this can be turned off if there's some exceptional reason why remoting3 tracing must still exist on the same classpath.

tracing/src/main/java/com/palantir/tracing/Tracer.java
qinfchen: i think this is more readable. also `remotingTracerVer.get()` is used to construct the error message, it makes sense to only do that when needed.

iamdanfox:
Closing out as I think we've got over the http-remoting -> conjure-java-runtime hump now!

PR #15, opened 2018-10-01
Tracing-aware functions, bifunctions, suppliers, consumers, biconsumers

mattsills:


iamdanfox:
Out of curiosity, can you paste a code snippet that you're intending to use these new `Tracers.wrap` methods?

mattsills:
@iamdanfox anything involving chaining futures.
```
CompletableFuture.supplyAsync(..., executor).thenApply(res -> ...);
```
```
CompletableFuture.supplyAsync(..., executor).thenApply(trace(res -> ...));
```

pkoenig10:
This won't actually work if you try to create a `Supplier`.

For example, this code:
```
Callable<Integer> callable = Tracers.wrap(() -> 42);
Supplier<Integer> supplier = Tracers.wrap(() -> 42);
```

will fail with the following errors:
> reference to wrap is ambiguous
 
> incompatible types: no instance(s) of type variable(s) V exist so that Callable<V> conforms to Supplier<Integer>

dansanduleac:
@mattsills do you still need this?
Also, @pkoenig10 raised a point that this breaks, can you comment on whether that's the case or not?

PR #22, opened 2018-10-25
Trace thread state is removed when root spans are completed

carterkozak:
Multiple root spans are not allowed, traceId state should not
leak between multiple root spans.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -46,11 +46,7 @@ public final class Tracer {
     private Tracer() {}
 
     // Thread-safe since thread-local
-    private static final ThreadLocal<Trace> currentTrace = ThreadLocal.withInitial(() -> {
-        Trace trace = createTrace(Optional.empty(), Tracers.randomId());
-        MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
-        return trace;
-    });
+    private static final ThreadLocal<Trace> currentTrace = new ThreadLocal<>();
 
     // Only access in a class-synchronized fashion
     private static final Map<String, SpanObserver> observers = new HashMap<>();
@@ -85,7 +81,8 @@ public final class Tracer {
      * when the current trace is empty.
      */
     public static OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
-        Preconditions.checkState(currentTrace.get().isEmpty(),
+        Trace current = getOrCreateCurrentTrace();
+        Preconditions.checkState(current.isEmpty(),
                 "Cannot start a span with explicit parent if the current thread's trace is non-empty");
         Preconditions.checkArgument(parentSpanId != null && !parentSpanId.isEmpty(),
                 "parentTraceId must be non-empty: %s", parentSpanId);
@@ -95,7 +92,7 @@ public final class Tracer {
                 .parentSpanId(parentSpanId)
                 .type(type)
                 .build();
-        currentTrace.get().push(span);
+        current.push(span);
         return span;
     }
 
@@ -119,13 +116,14 @@ public final class Tracer {
                 .spanId(Tracers.randomId())
                 .type(type);
 
-        Optional<OpenSpan> prevState = currentTrace.get().top();
+        Trace trace = getOrCreateCurrentTrace();
+        Optional<OpenSpan> prevState = trace.top();
         if (prevState.isPresent()) {
             spanBuilder.parentSpanId(prevState.get().getSpanId());
         }
 
         OpenSpan span = spanBuilder.build();
-        currentTrace.get().push(span);
+        trace.push(span);
         return span;
     }
 
@@ -143,10 +141,12 @@ public final class Tracer {
      * Like {@link #fastCompleteSpan()}, but adds {@code metadata} to the current span being completed.
      */
     public static void fastCompleteSpan(Map<String, String> metadata) {
-        popCurrentSpan()
-                .filter(openSpan -> currentTrace.get().isObservable())
-                .map(openSpan -> toSpan(openSpan, metadata))
-                .ifPresent(Tracer::notifyObservers);
+        if (isTraceObservable()) {
+            String traceId = getTraceId();
+            popCurrentSpan()
+                    .map(openSpan -> toSpan(openSpan, metadata, traceId))
+                    .ifPresent(Tracer::notifyObservers);
+        }
     }
 
     /**
@@ -161,15 +161,15 @@ public final class Tracer {
      * Like {@link #completeSpan()}, but adds {@code metadata} to the current span being completed.
      */
     public static Optional<Span> completeSpan(Map<String, String> metadata) {
+        boolean observable = isTraceObservable();
+        Trace trace = currentTrace.get();
         Optional<Span> maybeSpan = popCurrentSpan()
-                .map(openSpan -> toSpan(openSpan, metadata));
+                .map(openSpan -> toSpan(openSpan, metadata, trace.getTraceId()));
 
         // Notify subscribers iff trace is observable
-        maybeSpan.ifPresent(span -> {
-            if (currentTrace.get().isObservable()) {
-                notifyObservers(span);
-            }
-        });
+        if (maybeSpan.isPresent() && observable) {
+            notifyObservers(maybeSpan.get());
+        }
 
         return maybeSpan;
     }
@@ -181,12 +181,20 @@ public final class Tracer {
     }
 
     private static Optional<OpenSpan> popCurrentSpan() {
-        return currentTrace.get().pop();
+        Trace trace = currentTrace.get();
+        if (trace != null) {
+            Optional<OpenSpan> span = trace.pop();
+            if (trace.isEmpty()) {
+                clearCurrentTrace();
+            }
+            return span;
+        }
+        return Optional.empty();
     }
 
-    private static Span toSpan(OpenSpan openSpan, Map<String, String> metadata) {
+    private static Span toSpan(OpenSpan openSpan, Map<String, String> metadata, String traceId) {
         return Span.builder()
-                .traceId(getTraceId())
+                .traceId(traceId)
                 .spanId(openSpan.getSpanId())
                 .type(openSpan.type())
                 .parentSpanId(openSpan.getParentSpanId())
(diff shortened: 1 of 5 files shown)


carterkozak:
@schlosna fysa

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -239,14 +247,13 @@ public static void setSampler(TraceSampler sampler) {
 
     /** Returns the globally unique identifier for this thread's trace. */
     public static String getTraceId() {
-        return currentTrace.get().getTraceId();
+        return Preconditions.checkNotNull(currentTrace.get(), "There is no root span").getTraceId();
```
carterkozak, line 261: I considered returning a newly generated ID when no trace is available, but that could potentially result in bad data if we initialize a new span assuming that traceId will be maintained.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -46,11 +46,7 @@
     private Tracer() {}
 
     // Thread-safe since thread-local
-    private static final ThreadLocal<Trace> currentTrace = ThreadLocal.withInitial(() -> {
-        Trace trace = createTrace(Optional.empty(), Tracers.randomId());
-        MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
-        return trace;
-    });
+    private static final ThreadLocal<Trace> currentTrace = new ThreadLocal<>();
```
iamdanfox, line 49: OK so I guess this is the key change - this guy now uses `null` to represent the idea that we're not in a trace?

I guess we just have to be super careful now everywhere that we use currentTrace.get() that it might actually be null!

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -143,10 +141,12 @@ public static void fastCompleteSpan() {
      * Like {@link #fastCompleteSpan()}, but adds {@code metadata} to the current span being completed.
      */
     public static void fastCompleteSpan(Map<String, String> metadata) {
-        popCurrentSpan()
-                .filter(openSpan -> currentTrace.get().isObservable())
-                .map(openSpan -> toSpan(openSpan, metadata))
-                .ifPresent(Tracer::notifyObservers);
+        if (isTraceObservable()) {
```
iamdanfox: Why put this behind the `isTraceObservable` check?

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: That is correct, otherwise trace thread state leaks outside of root spans, which causes multiple root spans unless code directly creates a new traceId.

tracing/src/main/java/com/palantir/tracing/Tracer.java
iamdanfox: Possibly want to use the safelogging Preconditions here so these messages will definitely not get redacted?

https://github.com/palantir/safe-logging#preconditions

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: ack, this is a bug. fix incoming

iamdanfox:
I think this makes sense directionally, but I want to be super careful changing this stuff because it's been stable for quite a long time and it underpins pretty much every one of our services!

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -274,4 +286,17 @@ static void setTrace(Trace trace) {
...
+    private static Trace getOrCreateCurrentTrace() {
+        Trace trace = currentTrace.get();
+        if (trace == null) {
+            trace = createTrace(Optional.empty(), Tracers.randomId());
+            setTrace(trace);
+        }
+        return trace;
+    }
+
+    private static void clearCurrentTrace() {
```
iamdanfox, line 309: Looks like the only two places this can be called are:

- `public Trace getAndClearTrace()`
- when a call to `fastCompleteSpan()` -> `popCurrentSpan()` (pops the last remaining span)

carterkozak: That is correct

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: I agree that would be ideal, but we don't currently depend on safe-logging. I'd prefer not to add deps in this change.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: fixed

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -119,13 +116,14 @@ private static OpenSpan startSpanInternal(String operation, SpanType type) {
                 .spanId(Tracers.randomId())
                 .type(type);
 
-        Optional<OpenSpan> prevState = currentTrace.get().top();
+        Trace trace = getOrCreateCurrentTrace();
+        Optional<OpenSpan> prevState = trace.top();
         if (prevState.isPresent()) {
             spanBuilder.parentSpanId(prevState.get().getSpanId());
         }
```
schlosna, line 124: nit: could simplify

```suggestion
        trace.top().ifPresent(openSpan -> spanBuilder.parentSpanId(openSpan.getSpanId()));
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -143,10 +141,15 @@ public static void fastCompleteSpan() {
...
     public static void fastCompleteSpan(Map<String, String> metadata) {
-        popCurrentSpan()
-                .filter(openSpan -> currentTrace.get().isObservable())
-                .map(openSpan -> toSpan(openSpan, metadata))
-                .ifPresent(Tracer::notifyObservers);
+        Trace trace = currentTrace.get();
+        if (trace != null) {
+            boolean observable = isTraceObservable();
+            Optional<OpenSpan> span = popCurrentSpan();
+            if (observable) {
```
schlosna: should we reuse the trace to avoid an additional thread local get?

```suggestion
            if (isObservable(trace)) {
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -161,15 +164,15 @@ public static void fastCompleteSpan(Map<String, String> metadata) {
...
-                .map(openSpan -> toSpan(openSpan, metadata));
+                .map(openSpan -> toSpan(openSpan, metadata, trace.getTraceId()));
 
         // Notify subscribers iff trace is observable
-        maybeSpan.ifPresent(span -> {
-            if (currentTrace.get().isObservable()) {
-                notifyObservers(span);
-            }
-        });
+        if (maybeSpan.isPresent() && observable) {
```
schlosna: should we reuse the trace to avoid additional thread local get and simplify?

```suggestion
        if (isObservable(trace)) {
            maybeSpan.ifPresent(Tracer::notifyObservers);
        }
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -255,12 +265,17 @@ public static Trace getAndClearTrace() {
      * Tracer#completeSpan span completion}.
      */
     public static boolean isTraceObservable() {
-        return currentTrace.get().isObservable();
+        Trace trace = currentTrace.get();
+        return trace != null && trace.isObservable();
```
schlosna, line 277: related to above comments, could add `isObservable(Trace)` to avoid additional thread local lookups

```suggestion        
        return isObservable(currentTrace.get());
    }

    private static boolean isObservable(Trace trace) {
        return trace != null && trace.isObservable();
    }
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
schlosna: should we expose a `public static boolean hasTraceId()` to allow checking if tracing is currently enabled?

Once this PR goes through we can obviously clean up https://github.com/palantir/tritium/blob/develop/tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java#L71 as `getAndClearTrace` will now do the right thing and clear state, but there may be cases where we want to know if there is a root trace ID already set.

schlosna (commented):
Supportive of this PR as it will allow removing the workarounds added in https://github.com/palantir/tritium/pull/134

Have a few questions/nits below

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: I opted to avoid the lambda allocation since this is a relatively hot path. We do a lot elsewhere, so I could go either way.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: good call, I think we want `trace.isObservable()` rather than a new method unless I'm misunderstanding.

carterkozak: All the call sites should have a separate null check on the trace, so I think this will be okay.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: 👍

tracing/src/main/java/com/palantir/tracing/Tracer.java
schlosna: makes sense

carterkozak:
> should we expose a public static boolean hasTraceId() to allow checking if tracing is currently enabled?

I think that would be helpful, but I would prefer to propose that in a separate change.

uschi2000:
I don't see a test for the changed behavior, could you add one, please?

carterkozak:
Was tested (though admittedly indirectly) by this assertion: `assertThat(MDC.get(Tracers.TRACE_ID_KEY)).isNull();`

Added a better test for this specific change.

tracing/src/main/java/com/palantir/tracing/Tracer.java
iamdanfox: +1 to a `hasTraceId` or a new method to return `Optional<String>`... otherwise you can't eyeball I method like this and be confident it's safe:

```java

    @Override
    public HttpConnection create(URL url) throws IOException {
        HttpConnection connection = delegate.create(url);
        String traceId = Tracer.getTraceId();
        connection.setRequestProperty(TraceHttpHeaders.TRACE_ID, traceId);
        return connection;
    }
```

carterkozak: Added.

I searched for usages of getTraceId, and everything that I found called it after creating a new span, so it should be safe, though there may be cases that I'm not aware of. For what it's worth, anything that does throw here is also responsible for leaking traceIds.

iamdanfox (approved):
Overall, I think this change looks good.  It forces users to be a bit more precise rather than just assuming a root span will be invented.

I've also done a quick search internally for any users of the `getTraceId` method and it seems that very few people use it directly.

PR #27, opened 2018-10-29
Implement Trace.toString

carterkozak:


ferozco (approved):

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -81,4 +81,9 @@ String getTraceId() {
     Trace deepCopy() {
         return new Trace(new ArrayDeque<>(stack), isObservable, traceId);
     }
+
+    @Override
+    public String toString() {
+        return "Trace{stack=" + stack + ", isObservable=" + isObservable + ", traceId='" + traceId + "'}";
```
iamdanfox, line 87: does `Deque<String>` give you a helpful debug representation? I wouldn't want to just give people `Dequeu@cd57a67`.

Could we add a unit test just to demo the functionality?

carterkozak: Yep! https://github.com/palantir/tracing-java/pull/48

PR #31, opened 2018-10-31
fix #30: Sampling is disabled by default

### LINKED ISSUE #30: The default sampler should not sample all operations
## What happened?

If I create a simple application and include http-remoting, every requests is sent with `X-B3-Sampled: 1` due to our usage of `AlwaysSampler` by default. Even if my demonstration application does not write trace data to disk, it will cause other services to do so.

I am proposing that we update the default sampler to `new RandomSampler(.01f)` to sample 1% of actions unless another sampler is specified. I have no strong opinion regarding 1% in particular, but would prefer that we do not sample everything. This value is used elsewhere, so it should be a reasonable default.

carterkozak:
Previously the default configuration sampled every operation.
Any application using a tracing-aware client would send every
request with `X-B3-Sampled: 1`, so even if the local application
did not record sample data to disk, upstream services would
be forced to.


carterkozak:
Closing, we should pursue a proper fix leveraging an undecided state.

PR #45, opened 2018-11-27
Failing streaming endpoints do return trace IDs

buffer51:
## Before this PR
A failing streaming endpoint would not return the trace ID as a response header.

## After this PR
Currently this PR doesn't fix the issue, it just adds unit tests that show the problem.

### DIFF
--- a/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
+++ b/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
@@ -88,6 +88,9 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
             Span span = maybeSpan.get();
             headers.putSingle(TraceHttpHeaders.TRACE_ID, span.getTraceId());
         } else {
+            // When the filter is called twice (e.g. an exception is thrown in a streaming call),
+            // the current trace will be empty. To allow clients to still get the trace ID corresponding to
+            // the failure, we retrieve it from the requestContext.
             Optional.ofNullable(requestContext.getProperty(TRACE_ID_PROPERTY_NAME))
                     .ifPresent(s -> headers.putSingle(TraceHttpHeaders.TRACE_ID, s));
         }
--- a/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
+++ b/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
@@ -87,6 +87,9 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
         if (maybeSpan.isPresent()) {
             Span span = maybeSpan.get();
             headers.putSingle(TraceHttpHeaders.TRACE_ID, span.getTraceId());
+        } else {
+            Optional.ofNullable(requestContext.getProperty(TRACE_ID_PROPERTY_NAME))
+                    .ifPresent(s -> headers.putSingle(TraceHttpHeaders.TRACE_ID, s));
         }
     }
 
--- a/tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
+++ b/tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
@@ -49,6 +49,7 @@ import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 import org.glassfish.jersey.client.JerseyClientBuilder;
 import org.junit.After;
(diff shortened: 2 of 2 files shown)


buffer51:
The issue seems to be that the filter is called twice:
- first for the initial response, where the trace is set, but the response is eventually discarded because of the exception
- then a second response is created based on the exception alone, which goes to the filter but the trace has been cleared and nothing is set.

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
```diff
@@ -156,6 +157,36 @@ public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeade
         assertThat(spanCaptor.getValue().getOperation(), is("GET /trace"));
     }
 
+    @Test
+    public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeadersWhenFailing() {
+        Response response = target.path("/failing-trace").request().get();
+        assertThat(response.getHeaderString(TraceHttpHeaders.TRACE_ID), not(nullValue()));
```
uschi2000, line 163: taking a step back, it's not clear to me that the response needs to contain a trace id header at all (even in the success case).

buffer51: In the general case of streaming one? I'm confused, this is how it's been working so far everywhere. If not provided on the request headers, a trace will be created and returned so that you can find out what happens. This makes even more sense IMHO when things go wrong.

For what it's worth, here even forcing the trace with request headers is ignored in the failing streaming case. Doesn't really matter if you force it or let it pick, the exception handling breaks it.

uschi2000: cool, that sounds reasonable.

buffer51:
I don't know if that best effort else condition I put introduces some bad behaviour in other cases. I suspect it might, otherwise why wouldn't the logic be as simple as that? Perhaps because people don't only use automatic tracing, but the more advanced capabilities.

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -87,6 +87,9 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
         if (maybeSpan.isPresent()) {
             Span span = maybeSpan.get();
             headers.putSingle(TraceHttpHeaders.TRACE_ID, span.getTraceId());
+        } else {
```
uschi2000, line 90: It's pretty funky that this filter is called twice in the case of a failing streaming response... indicates to me that we don't fully understand the jersey behavior here.

ignoring our ignorance for a moment, the fix seems ok to me, but i'd like to see a comment indicating why this `else` block exists and what it does and how.

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
```diff
@@ -156,6 +157,36 @@ public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeade
         assertThat(spanCaptor.getValue().getOperation(), is("GET /trace"));
     }
 
+    @Test
+    public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeadersWhenFailing() {
```
uschi2000, line 161: note that this test succeeded before this fix

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
```diff
@@ -156,6 +157,36 @@ public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeade
...
+        Response response = target.path("/failing-trace").request().get();
+        assertThat(response.getHeaderString(TraceHttpHeaders.TRACE_ID), not(nullValue()));
+        assertThat(response.getHeaderString(TraceHttpHeaders.PARENT_SPAN_ID), is(nullValue()));
+        assertThat(response.getHeaderString(TraceHttpHeaders.SPAN_ID), is(nullValue()));
+        verify(observer).consume(spanCaptor.capture());
+        assertThat(spanCaptor.getValue().getOperation(), is("GET /failing-trace"));
+    }
+
+    @Test
+    public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeadersWhenStreaming() {
```
uschi2000, line 171: note that this test succeeded before this fix

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
```diff
@@ -156,6 +157,36 @@ public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeade
...
+        Response response = target.path("/streaming-trace").request().get();
+        assertThat(response.getHeaderString(TraceHttpHeaders.TRACE_ID), not(nullValue()));
+        assertThat(response.getHeaderString(TraceHttpHeaders.PARENT_SPAN_ID), is(nullValue()));
+        assertThat(response.getHeaderString(TraceHttpHeaders.SPAN_ID), is(nullValue()));
+        verify(observer).consume(spanCaptor.capture());
+        assertThat(spanCaptor.getValue().getOperation(), is("GET /streaming-trace"));
+    }
+
+    @Test
+    public void testTraceState_withoutRequestHeadersGeneratesValidTraceResponseHeadersWhenFailingToStream() {
```
uschi2000, line 181: this test fails before this fix, and succeeds with this fix

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
buffer51: Agreed, will do

uschi2000 (approved):
LGTM. @iamdanfox ?

iamdanfox (approved):
Thanks @buffer51!

PR #49, opened 2018-12-20
[fix] Open a new span when wrapping with a new trace

### LINKED ISSUE #33: Creating a new trace should always open a new span
It should not be possible to create a new trace without also creating a root span. The current API allows for an undefined gap between creating a trace, and opening a span. In this gap, a `traceId` is set on the thread (and logged messages) without occurring inside of an observed operation.

pkoenig10:
## Before this PR
Creating a new trace does not create a new root span. Thus all top-level calls in this trace will have no parent span. This breaks assumptions of trace log consumers who assume that there is only one root span for a given trace.

This also seems counter to the `parentId` specification in the [Zipkin API](https://github.com/openzipkin/zipkin-api/blob/5fd719b114ff2ba44f1d2757888fa123cdfb0da6/zipkin-api.yaml#L353):
> The parent span ID or absent if this the root span in a trace.

## After this PR
Creating a new trace will create a new root span.

The issue is part of the `Major API revision` milestone, but this fix does not break the API. Is a major API revision required for this change?

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ThreadLocalRandom;
 public final class Tracers {
     /** The key under which trace ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
     public static final String TRACE_ID_KEY = "traceId";
+    private static final String ROOT_SPAN_OPERATION = "root";
     private static final char[] HEX_DIGITS =
             {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
 
@@ -148,8 +149,10 @@ public final class Tracers {
 
             try {
                 Tracer.initTrace(Optional.empty(), Tracers.randomId());
+                Tracer.startSpan(ROOT_SPAN_OPERATION);
                 return delegate.call();
             } finally {
+                Tracer.completeSpan();
                 // restore the trace
                 Tracer.setTrace(originalTrace);
             }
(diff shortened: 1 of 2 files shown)


uschi2000 (approved):
lgtm

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -148,8 +149,10 @@ public static ScheduledExecutorService wrapWithNewTrace(ScheduledExecutorService
 
             try {
                 Tracer.initTrace(Optional.empty(), Tracers.randomId());
+                Tracer.startSpan(ROOT_SPAN_OPERATION);
                 return delegate.call();
             } finally {
+                Tracer.completeSpan();
```
carterkozak: fastCompleteSpan?

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -166,8 +169,10 @@ public static Runnable wrapWithNewTrace(Runnable delegate) {
 
             try {
                 Tracer.initTrace(Optional.empty(), Tracers.randomId());
+                Tracer.startSpan(ROOT_SPAN_OPERATION);
                 delegate.run();
             } finally {
+                Tracer.completeSpan();
```
carterkozak: fastCompleteSpan

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -148,8 +149,10 @@ public static ScheduledExecutorService wrapWithNewTrace(ScheduledExecutorService
```
carterkozak, line 149: Looks like there's an existing bug where wrapping a callable will leave an empty trace with no root span on the thread if there's not already an active trace.

`Tracer.getAndClearTrace` should probably return `Optional<Trace>`

Not related to this change though.

carterkozak:
> The issue is part of the Major API revision milestone, but this fix does not break the API. Is a major API revision required for this change?

API rev isn't required for this change, but I don't think it solves the linked ticket. This definitely fixes a bug (thanks!), but it's still possible to invoke `Tracer.initTrace` without also creating a root span, which is what I'd like to target in a major version rev.

carterkozak (approved):

Issue #51, opened 2019-01-10
Call to wrapWithNewTrace incorrectly restores state if there is no current trace


dsd987:
## What happened?
Calling Tracers.wrapWithNewTrace when no trace is currently set will result in a new trace being restored on completion rather than a cleared trace state.

## What did you want to happen?
The trace state should be restored back to the same state as before the call was made, so in this case, a cleared trace state.

carterkozak:
Fixed in 2.1.1

PR #52, opened 2019-01-10
Restore trace state back to cleared when wrapWithNewTrace is called with no trace state

dsd987:
Previously, when `Tracers::wrapWithNewTrace` was called with no trace state, the trace state would incorrectly be restored to a new trace rather than the cleared state. Now, the trace state will be appropriately returned to the cleared state.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -270,6 +270,13 @@ public final class Tracer {
         return Preconditions.checkNotNull(currentTrace.get(), "There is no root span").getTraceId();
     }
 
+    /** Clears the current trace id and returns it if present. */
+    static Optional<Trace> getAndClearTraceIfPresent() {
+        Optional<Trace> trace = Optional.ofNullable(currentTrace.get());
+        clearCurrentTrace();
+        return trace;
+    }
+
     /** Clears the current trace id and returns (a copy of) it. */
     public static Trace getAndClearTrace() {
         Trace trace = getOrCreateCurrentTrace();
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -145,7 +145,7 @@ public final class Tracers {
     public static <V> Callable<V> wrapWithNewTrace(Callable<V> delegate) {
         return () -> {
             // clear the existing trace and keep it around for restoration when we're done
-            Trace originalTrace = Tracer.getAndClearTrace();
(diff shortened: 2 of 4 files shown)


dsd987:
Proposed fix for #51

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -270,6 +270,13 @@ public static String getTraceId() {
         return Preconditions.checkNotNull(currentTrace.get(), "There is no root span").getTraceId();
     }
 
+    /** Clears the current trace id and returns (a copy of) it if present */
```
carterkozak: I don't think this makes a copy

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -188,20 +186,32 @@ public static Runnable wrapWithNewTrace(Runnable delegate) {
...
+     */
+    private static void restoreTrace(Optional<Trace> trace) {
+        if (trace.isPresent()) {
+            Tracer.setTrace(trace.get());
+        } else {
+            // Ignoring returned value, used to clear trace only
+            Tracer.getAndClearTraceIfPresent();
+        }
+    }
+
```
carterkozak, line 214: needs tests!

tracing/src/test/java/com/palantir/tracing/TracerTest.java
```diff
@@ -251,6 +251,20 @@ public void testFastCompleteSpanWithMetadata() {
...
     }
 
+    @Test
+    public void testGetAndClearTraceIfPresent() {
+        Trace trace = new Trace(true, "newTraceId");
+        Tracer.setTrace(trace);
+
+        Optional<Trace> nonEmptyTrace = Tracer.getAndClearTraceIfPresent();
+        assertThat(nonEmptyTrace).isPresent();
+        assertThat(nonEmptyTrace).hasValue(trace);
```
carterkozak, line 260: isPresent isn't necessary if we're checking it has a specific value

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -270,6 +270,13 @@ public static String getTraceId() {
         return Preconditions.checkNotNull(currentTrace.get(), "There is no root span").getTraceId();
     }
 
+    /** Clears the current trace id and returns it if present. */
+    static Optional<Trace> getAndClearTraceIfPresent() {
```
carterkozak, line 274: I'd like if this was public, but we can look into that in a separate PR!

carterkozak (approved):
Thanks!

PR #58, opened 2019-01-26
[improvement] Distinguish Jersey and OkHttp operation names

markelliot:
## Before this PR
Cannot identify whether a span originates from OkHttp or Jersey.

## After this PR
Can identify where a span originates.


j-baker (approved):

dansanduleac:
Can't we use the span type to differentiate whether it's a client side or server side span?

PR #59, opened 2019-01-29
[improvement] Add methods to set span operation when wrapping with a new trace

pkoenig10:
## Before this PR
When calling any of the `wrapWithNewTrace` methods, the operation of the initial span would be set to "root". This makes it difficult to determine the origin of traces if `wrapWithNewTrace` is used in more than one place.

## After this PR
`Tracers` now provides variants of the `wrapWithNewTrace` methods that allow callers to explicitly set the initial span operation.

Follow up to #49.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -105,51 +105,73 @@ public final class Tracers {
     }
 
     /**
+     * Like {@link #wrapWithNewTrace(String, ExecutorService)}, but with a default initial span operation.
+     */
+    public static ExecutorService wrapWithNewTrace(ExecutorService executorService) {
+        return wrapWithNewTrace(ROOT_SPAN_OPERATION, executorService);
+    }
+
+    /**
      * Wraps the provided executor service to make submitted tasks traceable with a fresh {@link Trace trace}
-     * for each execution, see {@link #wrapWithNewTrace(ExecutorService)}. This method should not be used to
+     * for each execution, see {@link #wrapWithNewTrace(String, ExecutorService)}. This method should not be used to
      * wrap a ScheduledExecutorService that has already been {@link #wrap(ExecutorService) wrapped}. If this is
      * done, a new trace will be generated for each execution, effectively bypassing the intent of the previous
      * wrapping.
      */
-    public static ExecutorService wrapWithNewTrace(ExecutorService executorService) {
+    public static ExecutorService wrapWithNewTrace(String operation, ExecutorService executorService) {
         return new WrappingExecutorService(executorService) {
             @Override
             protected <T> Callable<T> wrapTask(Callable<T> callable) {
-                return wrapWithNewTrace(callable);
+                return wrapWithNewTrace(operation, callable);
             }
         };
     }
 
     /**
-     * Wraps the provided scheduled executor service to make submitted tasks traceable with a fresh {@link Trace trace}
-     * for each execution, see {@link #wrapWithNewTrace(ScheduledExecutorService)}. This method should not be used to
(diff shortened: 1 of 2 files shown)


tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -237,7 +240,7 @@ public void testWrapCallableWithNewTrace_traceStateInsideCallableHasSpan() throw
 
         OpenSpan span = spans.get(0);
 
-        assertThat(span.getOperation()).isEqualTo("root");
```
carterkozak: we should retain test cases for the default as well

carterkozak:
It would be helpful to configure our executors to wrap operations based on the executor name.

This looks sane to me, thoughts @uschi2000?

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -104,52 +104,74 @@ public static Runnable wrap(Runnable delegate) {
...
     /**
      * Wraps the provided executor service to make submitted tasks traceable with a fresh {@link Trace trace}
-     * for each execution, see {@link #wrapWithNewTrace(ExecutorService)}. This method should not be used to
+     * for each execution, see {@link #wrapWithNewTrace(String, ExecutorService)}. This method should not be used to
      * wrap a ScheduledExecutorService that has already been {@link #wrap(ExecutorService) wrapped}. If this is
      * done, a new trace will be generated for each execution, effectively bypassing the intent of the previous
      * wrapping.
      */
-    public static ExecutorService wrapWithNewTrace(ExecutorService executorService) {
+    public static ExecutorService wrapWithNewTrace(String operation, ExecutorService executorService) {
```
iamdanfox: This will appear unredacted in Telemetry, so what do you think about another beloved `@CompileTimeConstant` to enforce that people don't accidentally exfil usernames or something?

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -121,11 +122,12 @@ public void testScheduledExecutorServiceWrapsCallablesWithNewTraces() throws Exc
 
     @Test
     public void testExecutorServiceWrapsCallablesWithNewTraces() throws Exception {
+        String someOperartion = "operationToUse";
         ExecutorService wrappedService =
-                Tracers.wrapWithNewTrace(Executors.newSingleThreadExecutor());
+                Tracers.wrapWithNewTrace(someOperartion, Executors.newSingleThreadExecutor());
```
iamdanfox: someOperation?

tracing/src/main/java/com/palantir/tracing/Tracers.java
pkoenig10: This would make it impossible to use thread pool name patterns as the operation when created a wrapped executor.

carterkozak: I don't think this is the operation we'd use for thread pools, we'd need to use the DeferredTracer + overload Tracers.wrap(ExecutorService) which can be implemented in a separate PR

pkoenig10: We use this method when building scheduled executor services with new traces.

ellisjoe:
Does this mean that all of these traces will no longer have any `root` span? Is that valid given the zipkin spec?

carterkozak:
They won't have a span named `root`, but they will still have a root span. Normally our root span is something along the lines of `Jersey: GET /foo`.

Related to this PR, do we think we should add a span for every runnable executed on an executor? We copy the trace and allow spans to be created, but we don't provide any data to track when the task began/completed, or on which executor.

carterkozak:
This looks reasonable to me 👍

pkoenig10:
I've opened https://github.com/palantir/tracing-java/issues/63

PR #67, opened 2019-02-07
fix #66: SpanObserver doesn't prevent execution on failure

### LINKED ISSUE #66: SpanObserver failure should neither be propagated nor block other observers
Currently a throwing observer will block execution of the next, and the exception is propagated to the called. I would expect an error to be logged, but execution to be otherwise unaffected.

carterkozak:
An error is logged, and execution will continue, including other
SpanObservers registered to the Tracer.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -17,7 +17,6 @@
 package com.palantir.tracing;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.tracing.api.OpenSpan;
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -252,7 +250,23 @@ public static synchronized SpanObserver unsubscribe(String name) {
     }
 
     private static void computeObserversList() {
-        observersList = ImmutableList.copyOf(observers.values());
+        Consumer<Span> newCompositeObserver = span -> { };
```
iamdanfox, line 253: If we're micro-optimising, why do we tolerate calling the empty observer first and calling `andThen` on it?

carterkozak: it's just a method invocation which is free compared to anything that does allocation (list iterator, for example).

carterkozak: I can get rid of it, but I don't think it's worth the complexity since we won't be able to measure a difference.

iamdanfox (approved):

PR #70, opened 2019-02-08
Use logsafe preconditions

pkoenig10:
A null `currentTrace` means that there is no trace, not there is no root span. In the current implementation is is possible to have a trace but no root span.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -16,7 +16,10 @@
 
 package com.palantir.tracing;
 
-import com.google.common.base.Preconditions;
+import static com.palantir.logsafe.Preconditions.checkArgument;
+import static com.palantir.logsafe.Preconditions.checkNotNull;
+import static com.palantir.logsafe.Preconditions.checkState;
+
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.tracing.api.OpenSpan;
@@ -62,7 +65,9 @@ public final class Tracer {
      * configured sampler} returns true.
      */
     private static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
-        Preconditions.checkArgument(traceId != null && !traceId.isEmpty(), "traceId must be non-empty: %s", traceId);
+        checkArgument(traceId != null && !traceId.isEmpty(),
+                "traceId must be non-empty",
+                SafeArg.of("traceId", traceId));
         boolean observable = shouldObserve(isObservable);
         return new Trace(observable, traceId);
     }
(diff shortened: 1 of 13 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -281,7 +281,7 @@ public static boolean hasTraceId() {
 
     /** Returns the globally unique identifier for this thread's trace. */
     public static String getTraceId() {
-        return Preconditions.checkNotNull(currentTrace.get(), "There is no root span").getTraceId();
+        return Preconditions.checkNotNull(currentTrace.get(), "There is no trace").getTraceId();
```
carterkozak: Should we switch to safe-logging preconditions while we're at it?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -62,7 +65,9 @@ private Tracer() {}
      * configured sampler} returns true.
      */
     private static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
-        Preconditions.checkArgument(traceId != null && !traceId.isEmpty(), "traceId must be non-empty: %s", traceId);
+        checkArgument(traceId != null && !traceId.isEmpty(),
+                "traceId must be non-empty",
+                SafeArg.of("traceId", traceId));
```
carterkozak: traceId is either null or empty, I don't think this arg is terribly helpful

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -62,7 +65,9 @@ private Tracer() {}
      * configured sampler} returns true.
      */
     private static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
-        Preconditions.checkArgument(traceId != null && !traceId.isEmpty(), "traceId must be non-empty: %s", traceId);
+        checkArgument(traceId != null && !traceId.isEmpty(),
```
carterkozak: not your change, but we could use `!Strings.isNullOrEmpty(traceId)`

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -91,10 +96,11 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
...
     public static OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
         Trace current = getOrCreateCurrentTrace();
-        Preconditions.checkState(current.isEmpty(),
+        checkState(current.isEmpty(),
                 "Cannot start a span with explicit parent if the current thread's trace is non-empty");
-        Preconditions.checkArgument(parentSpanId != null && !parentSpanId.isEmpty(),
-                "parentTraceId must be non-empty: %s", parentSpanId);
+        checkArgument(parentSpanId != null && !parentSpanId.isEmpty(),
+                "parentSpanId must be non-empty",
+                SafeArg.of("parentSpanId", parentSpanId));
```
carterkozak: nit: Since this is a very hot code path, I'd prefer to rewrite checks with args to this form, avoiding Arg creation:

```java
if (parentSpanId == null || parentSpanId.isEmpty()) {
    throw new SafeIllegalArgumentException("parentSpanId must be non-empty",
                SafeArg.of("parentSpanId", parentSpanId));
}
```

carterkozak:
Sorry for the snowball of requests

tracing/src/main/java/com/palantir/tracing/Tracer.java
pkoenig10: Per your comment above, this arg isn't very helpful so we can probably just remove it and eliminate most of the cost here.

carterkozak (approved):
Thanks!

PR #71, opened 2019-02-08
Add methods to create wrapped executor with new span

### LINKED ISSUE #63: Create new span when wrapping executor tasks
When running a task in a wrapped executor, a new span should be created for the task. This will allow tracing information to clearly indicate when the task started and finished running in the executor.

See https://github.com/palantir/tracing-java/pull/59#issuecomment-461086792.

pkoenig10:
Fixes #63

~This PR preserves the existing executor wrapping methods and their behavior. So users must explicitly opt-in to the new span-creating behavior.~

This is no longer true. The existing executor methods will now create a new span for tasks they run.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -16,7 +16,6 @@
 
 package com.palantir.tracing;
 
-import com.google.errorprone.annotations.CompileTimeConstant;
 import java.util.Optional;
 
 /**
@@ -46,7 +45,7 @@ public final class DeferredTracer {
         this(Optional.empty());
     }
 
-    public DeferredTracer(@CompileTimeConstant String operation) {
+    public DeferredTracer(String operation) {
         this(Optional.of(operation));
     }
 
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -16,6 +16,7 @@
 
 package com.palantir.tracing;
 
+import com.google.errorprone.annotations.CompileTimeConstant;
 import java.util.Optional;
 
 /**
@@ -39,9 +40,19 @@ import java.util.Optional;
  */
 public final class DeferredTracer {
     private final Optional<Trace> trace;
+    private final Optional<String> operation;
 
     public DeferredTracer() {
+        this(Optional.empty());
+    }
+
+    public DeferredTracer(@CompileTimeConstant String operation) {
+        this(Optional.of(operation));
+    }
+
+    DeferredTracer(Optional<String> operation) {
         this.trace = Tracer.copyTrace();
+        this.operation = operation;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


diogoholanda:
What's the concern around making this the default behavior if it is generally useful? Are there any potential concerns?

pkoenig10:
Making it the default behavior would require changing the API to also take an operation.

I would prefer to deprecate the old methods. I just wanted to get some agreement that this is the desired outcome.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -146,57 +162,91 @@ public void testExecutorServiceWrapsCallablesWithNewTraces() throws Exception {
...
+
+    @Test
+    public void testWrapCallable_callableTraceIsIsolated() throws Exception {
         Tracer.startSpan("outside");
-        Runnable runnable = Tracers.wrap(new Runnable() {
-            @Override
-            public void run() {
-                Tracer.startSpan("inside"); // never completed
-            }
+        Callable<Void> runnable = Tracers.wrap(() -> {
```
dansanduleac: rename to callable

pkoenig10:
@dansanduleac how do you feel about deprecating the old methods?

pkoenig10:
An alternative is to keep the old methods but use the default "root" span operation. However this probably isn't very useful since seeing a "root" span operation can be confusing.

diogoholanda:
I would actually challenge that. I would suggest making the operation not configurable, and simply something like "executor-wrapped-task". If someone wants to have an operation name, they can always create a new span themselves.

In the end, what we are trying to get from this feature is the knowledge of when the task actually run, so the operation name does not really matter and it is better to everyone to have the better behaviour by default.

Does that make sense?

pkoenig10:
I disagree. I think having a default span name that will be used for all executors will quickly become confusing when you have more than one executor.

Ideally users will want to identify these spans. Requiring them to wrap a task twice just to get useful names (with no way to remove the unhelpful default name) does not seems like the right solution here.

diogoholanda:
Why do you need to identify the executor that wrapped a task?
Is the assumption that a single task can be wrapped by multiple different executors and you want to know which one wrapped a given execution of the task?
That seems somewhat contrived.

Can you give me an explicit workflow in which a constant operation is not enough?

All that said, if you really feel strongly, we could allow the operation name to be configurable, but allowing for a default constant one in case you did not opt-in for overriding it seems much more beneficial than giving you no span at all until you sign in for it.

diogoholanda:
I don't mind making the operation name configurable. I guess it can be useful if you want to search for spans of a specific wrap.
Just arguing here for a different behavior in case the operation name is not provided

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -39,9 +40,19 @@
  */
 public final class DeferredTracer {
     private final Optional<Trace> trace;
+    private final Optional<String> operation;
 
     public DeferredTracer() {
+        this(Optional.empty());
+    }
+
+    public DeferredTracer(@CompileTimeConstant String operation) {
```
j-baker: Please remove CompileTimeConstant

j-baker: (because it's reasonable for this to not be a compiletimeconstant, and it is overly onerous for downstream developers)

pkoenig10:
Thoughts on adding a default span operation for the existing executor methods? It seems like a reasonable approach to me I just want to make sure we're willing to make that behavior change.

diogoholanda:
I would welcome that

dansanduleac:
👍

PR #74, opened 2019-02-13
[fix] Remove @CompileTimeConstant

j-baker:
## Before this PR
@CompileTimeConstant was used.

## After this PR
In practice, it can be tricky to make sure that everything is a CompileTimeConstant. Rather than demanding this at the lowest levels of our stack, we should be enforcing appropriate behaviour higher up. In particular, we would like to provide this functionality by default in an internal library, but the compile-time nature of this cannot be guaranteed by error-prone through a chain of lambdas.

So, let's not make it be a @CompileTimeConstant.


carterkozak:
👍

PR #77, opened 2019-02-19
[improvement] Add support for tracing-aware Guava FutureCallbacks.

dsd987:
## Before this PR
Functionality defined as part of a Guava FutureCallback may not be correctly traced as the thread it runs on may not be the thread adding the callback.

## After this PR
FutureCallbacks can be wrapped such that their traces are parented to the thread adding the callback or wrapped in a new trace if desired.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -16,11 +16,13 @@
 
 package com.palantir.tracing;
 
+import com.google.common.util.concurrent.FutureCallback;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
+import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 
 /** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Tracers {
@@ -102,6 +104,11 @@ public final class Tracers {
     /** Like {@link #wrap(Callable)}, but for Runnables. */
     public static Runnable wrap(Runnable delegate) {
         return new TracingAwareRunnable(delegate);
+    }
+
+    /** Like {@link #wrap(Callable)}, but for Guava's FutureCallback. */
+    public static <V> FutureCallback<V> wrap(FutureCallback<V> delegate) {
+        return new TracingAwareFutureCallback<>(delegate);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -206,6 +213,50 @@ public static Runnable wrapWithNewTrace(String operation, Runnable delegate) {
...
+     * Like {@link #wrapWithNewTrace(Callable)}, but for Guava's FutureCallback.
+     */
+    public static <V> FutureCallback<V> wrapWithNewTrace(FutureCallback<V> delegate) {
+        return wrapWithNewTrace(ROOT_SPAN_OPERATION, delegate);
+    }
+
+    /**
+     * Like {@link #wrapWithNewTrace(String, Callable)}, but for Guava's FutureCallback.
+     */
+    public static <V> FutureCallback<V> wrapWithNewTrace(String operation, FutureCallback<V> delegate) {
```
carterkozak: I don't think we ever want to create a _new_ trace around just the callback

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -450,6 +457,292 @@ public void testWrapRunnableWithAlternateTraceId_traceStateRestoredToCleared() {
...
+                Tracer.startSpan("inside"); // never completed
+            }
+        };
+
+        ListeningExecutorService listeningExecutorService =
+                MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor());
+        ListenableFuture<Void> success = listeningExecutorService.submit(() -> null);
+        ListenableFuture<Void> failure = listeningExecutorService.submit(() -> {
+            throw new IllegalStateException();
+        });
```
carterkozak: should use separate test cases for the successful/failed futures. Otherwise when tests fail, it. takes additional debugging to figure out what the problem is.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -450,6 +457,292 @@ public void testWrapRunnableWithAlternateTraceId_traceStateRestoredToCleared() {
...
+                assertThat(Tracer.completeSpan().get().getOperation()).isEqualTo("before-construction");
+            }
+        };
+
+        ListeningExecutorService listeningExecutorService =
+                MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor());
+        ListenableFuture<Void> success = listeningExecutorService.submit(() -> null);
+        ListenableFuture<Void> failure = listeningExecutorService.submit(() -> {
+            throw new IllegalStateException();
+        });
```
carterkozak, line 542: Should use separate test cases for these

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -104,6 +106,11 @@ public static Runnable wrap(Runnable delegate) {
         return new TracingAwareRunnable(delegate);
     }
 
+    /** Like {@link #wrap(Callable)}, but for Guava's FutureCallback. */
+    public static <V> FutureCallback<V> wrap(FutureCallback<V> delegate) {
+        return new TracingAwareFutureCallback<>(delegate);
```
carterkozak: Should this take an operation name? Once #71 merges DeferredTracer can take an operation name. I'd prefer we we required an operation name for new methods, and did not provide a method using a default.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -450,6 +457,292 @@ public void testWrapRunnableWithAlternateTraceId_traceStateRestoredToCleared() {
...
+
+        Tracer.startSpan("before-construction");
+        FutureCallback<Void> successCallback = Tracers.wrap(futureCallback);
+        Tracer.startSpan("after-construction");
+        Futures.addCallback(success, successCallback, MoreExecutors.directExecutor());
```
carterkozak: We should have a test that runs on an executor using a different thread, since tracing-java heavily uses thread state. The direct executor just invokes `run()` on the current thread.

carterkozak (commented):
@ellisjoe is putting some thought into support for tracing asynchronous operations as well.

tracing/src/main/java/com/palantir/tracing/Tracers.java
dsd987: Sure. Can wait until #71 merges and then rebase to use the DeferredTracer with the operation name.

PR #78, opened 2019-02-26
[fix] Don't use deprecated Futures#addCallback method

robert3005:
## Before this PR
tracing-java doesn't work with guava 27

## After this PR
tracing-java works with guava 27


carterkozak (approved):
Thanks!

carterkozak:
We should lint for deprecation

PR #79, opened 2019-02-26
Lint deprecation

carterkozak:
Prevents issues like #78


tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -38,6 +38,7 @@
 import org.mockito.MockitoAnnotations;
 import org.slf4j.MDC;
 
+@SuppressWarnings("deprecation")
```
robert3005, line 41: where is the deprecation coming from here?

robert3005 (approved):
LGTM

tracing/src/test/java/com/palantir/tracing/TracersTest.java
carterkozak: Several tests for Tracers methods we have deprecated

iamdanfox (approved):

PR #83, opened 2019-02-28
Manually upgrade dependencies

carterkozak:


carterkozak:
@iamdanfox @dansanduleac Do we have excavator running against this repo?

iamdanfox:
:+1: (need to look into if excavator-oss is unhappy)

PR #85, opened 2019-03-01
When Tracer.completeSpan() result is unused, prefer fastCompleteSpan

carterkozak:
The latter avoids work when operations are not sampled, which is
significantly more common than sampling.

Added error-prone annotations to validate correct usage.

### DIFF
--- a/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
+++ b/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
@@ -56,7 +56,7 @@ public enum OkhttpTraceInterceptor implements Interceptor {
         try {
             response = chain.proceed(tracedRequest.build());
         } finally {
-            Tracer.completeSpan();
+            Tracer.fastCompleteSpan();
         }
 
         return response;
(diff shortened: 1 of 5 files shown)


tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
```diff
@@ -56,7 +56,7 @@ public Response intercept(Chain chain) throws IOException {
         try {
             response = chain.proceed(tracedRequest.build());
         } finally {
-            Tracer.completeSpan();
+            Tracer.fastCompleteSpan();
```
carterkozak, line 59: Only usage change in main sources

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -167,14 +168,18 @@ public static void fastCompleteSpan(Map<String, String> metadata) {
     /**
      * Completes and returns the current span (if it exists) and notifies all {@link #observers subscribers} about the
      * completed span.
+     * If the return value is not used, prefer {@link Tracer#fastCompleteSpan()}.
      */
+    @CheckReturnValue
```
iamdanfox, line 173: :+1: this is a pretty neat way of ensuring we don't do unused work :)

iamdanfox:
:+1:

PR #89, opened 2019-03-07
[improvement] Revert adding spans to all wrapped executor tasks

diogoholanda:
## Before this PR
In https://github.com/palantir/tracing-java/pull/71, all wrapped executor tasks had spans added by default. This was pretty noisy for tasks that were already doing their own wrapping.

## After this PR
If wrapping is not explicitly requested, it is not done


diogoholanda:
@dansanduleac @bavardage @pkoenig10 
I guess I was too optimistic in the previous PR comments

dansanduleac:
👍

PR #92, opened 2019-03-09
[feature] Implement LeakedTraceFilter to protect services from Trace state leaks

carterkozak:
Previously services that did not use tracing internally would
often report traceIds in their logging associated with other
requests. This filter prevents filtered components from leaking
themselves, as well as from other components which leak and may not
be protected by this filter.


tracing-servlet/src/main/java/com/palantir/tracing/servlet/TraceStateBarrierFilter.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        try {
+            chain.doFilter(request, response);
+        } finally {
+            if (Tracer.hasTraceId()) {
+                Trace leakedTrace = Tracer.getAndClearTrace();
+                log.warn("This operation has leaked Tracer state. Tracer.startSpan was executed without "
+                                + "Tracer.completeSpan, resulting in both loss of span data and spans using "
+                                + "completion information from incorrect operations. Trace: {}",
+                        traceArg(leakedTrace));
```
schlosna: is there any other information we should capture as part of this (e.g. request path, headers, etc.?) to help track down what in the filter chain leaked?

carterkozak: I don't think we have access to any safe vars, but we can add the path as an unsafe parameter. Not unlikely that other MDC info has also been leaked here... but we shouldn't trust it all that much. TODO on me to write a similar filter in wc or sls-logging for a similar barrier for MDC.

iamdanfox:
:+1:

PR #95, opened 2019-03-12
[improvement] Add a utility class similar to `DeferredTracer` which supports (java) serialization

bavardage:
## Before this PR
I want to have traces span work done on worker nodes running in separate jvms (e.g. spark executors). I could previously have captured the traceid and then used `Tracers.wrapWithAlternateTraceId(...)` but this loses parent span info/isObservable.

## After this PR
I can create a `SerializableDeferredTracer` on the master node and have that be serialized out to the worker nodes.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/SerializableDeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/SerializableDeferredTracer.java
@@ -1,0 +1,105 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


dansanduleac (approved):

tracing/src/main/java/com/palantir/tracing/SerializableDeferredTracer.java
```diff
@@ -0,0 +1,105 @@
...
+ *     return null;
+ * });
+ *
+ * </code>
+ * </pre>
+ *
+ * N.b. the captured trace is restored without the full stack of spans, and so it's not possible to complete spans
+ * not started within the deferred context.
+ */
+public class SerializableDeferredTracer implements Serializable {
```
carterkozak, line 49: missing a serializable version id

carterkozak (changes requested):
Why is this preferable over updating `DeferredTracer` to be serializable?

bavardage:
not sure if desirable, but `DeferredTracer` maintains the full stack of spans

carterkozak:
If that's required for functionality, shouldn't this implementation do the same?
If it's not required for functionality, why should DeferredTracer maintain it?

bavardage:
closing for #97

PR #96, opened 2019-03-12
[improvement] wrapWithAlternateTraceId accepts observable state

diogoholanda:
## Before this PR
wrapWithAlternateTraceId did not accept the observable state to use. That means that a trace that was originally observable would not be fully observed when being propagated through this method.
An example of when this would be an issue if when propagating a trace from a Spark driver to a Spark executor, for instance by implementing a `ForeachPartitionFunction`.

## After this PR
The observable state can be properly propagated via wrapWithAlternateTraceId


diogoholanda:
Closing this since we seem to be preferring making DeferredTracer java serializable

PR #97, opened 2019-03-13
make `DeferredTracer` serializable

bavardage:
alternate approach to #95 

## Before this PR
I want to have traces span work done on worker nodes running in separate jvms (e.g. spark executors). I could previously have captured the traceid and then used Tracers.wrapWithAlternateTraceId(...) but this loses parent span info/isObservable.

## After this PR
I can create a `DeferredTracer` on the master node and have that be serialized out to the worker nodes.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -52,41 +52,37 @@ public final class DeferredTracer implements Serializable {
 
     private final String traceId;
     private final boolean isObservable;
+    @Nullable
     private final String operation;
     @Nullable
     private final String parentSpanId;
 
-    /** Create a new deferred tracer without specifying an operation.
-     * @deprecated use {@link #DeferredTracer(String)}.
-     */
-    @Deprecated
     public DeferredTracer() {
-        this(UNKNOWN_OPERATION);
-    }
-
-    /** Create a new deferred tracer, optionally specifying an operation.
-     * @deprecated use {@link #DeferredTracer(String)}.
-     */
-    @Deprecated
-    public DeferredTracer(Optional<String> operation) {
-        this(operation.orElse(UNKNOWN_OPERATION));
+        this(Optional.empty());
     }
 
     public DeferredTracer(String operation) {
+        this(Optional.of(operation));
+    }
+
+    /**
+     * Create a new deferred tracer, optionally specifying an operation.
+     * If no operation is specified, will attempt to use the parent span's operation name.
+     */
+    public DeferredTracer(Optional<String> operation) {
         Optional<Trace> maybeTrace = Tracer.copyTrace();
         if (maybeTrace.isPresent()) {
             Trace trace = maybeTrace.get();
             this.traceId = trace.getTraceId();
             this.isObservable = trace.isObservable();
             this.parentSpanId = trace.top().map(OpenSpan::getSpanId).orElse(null);
-
+            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(UNKNOWN_OPERATION));
         } else {
             this.traceId = null;
             this.isObservable = false;
             this.parentSpanId = null;
+            this.operation = null;
         }
-
-        this.operation = operation;
     }
 
     /**
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -48,8 +48,9 @@ public final class DeferredTracer implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    private static final String UNKNOWN_OPERATION = "deferred";
+    private static final String DEFAULT_OPERATION = "deferred";
 
+    @Nullable
     private final String traceId;
     private final boolean isObservable;
     @Nullable
@@ -76,7 +77,7 @@ public final class DeferredTracer implements Serializable {
             this.traceId = trace.getTraceId();
             this.isObservable = trace.isObservable();
             this.parentSpanId = trace.top().map(OpenSpan::getSpanId).orElse(null);
-            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(UNKNOWN_OPERATION));
+            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(DEFAULT_OPERATION));
         } else {
             this.traceId = null;
             this.isObservable = false;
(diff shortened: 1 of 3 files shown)


tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -34,23 +38,54 @@
...
     }
 
-    public DeferredTracer(String operation) {
-        this(Optional.of(operation));
+    /** Create a new deferred tracer, optionally specifying an operation.
+     * @deprecated use {@link #DeferredTracer(String)}.
+     */
+    @Deprecated
+    public DeferredTracer(Optional<String> operation) {
+        this(operation.orElse(UNKNOWN_OPERATION));
```
dansanduleac: how about attempting to capture the top span's operation, and use that?

bavardage: done

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -49,26 +65,48 @@ public DeferredTracer(String operation) {
...
+        }
     }
 
     /**
      * Runs the given callable with the current trace at
      * the time of construction of this {@link DeferredTracer}.
      */
     public <T, E extends Throwable> T withTrace(Tracers.ThrowingCallable<T, E> inner) throws E {
-        if (!trace.isPresent()) {
+        if (traceId == null) {
```
carterkozak, line 94: if we specify an operation name, but haven't started a trace yet, the deferred tracer will not create a new trace or span. Is that expected?

carterkozak: I suppose it matches the previous logic, so not something we should focus on in this PR.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -34,12 +38,24 @@
...
  * </pre>
  */
-public final class DeferredTracer {
-    private final Optional<Trace> trace;
-    private final Optional<String> operation;
+public final class DeferredTracer implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final String UNKNOWN_OPERATION = "deferred";
```
carterkozak: Does `DEFAULT_OPERATION` make more sense?

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -34,12 +38,24 @@
...
+
+    private final String traceId;
```
carterkozak, line 54: `@Nullable`

bavardage: whoops, done

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
bavardage: yeah

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
bavardage: yeah, not sure - it matches the previous behaviour

carterkozak:
Looks good to me, I like this approach a great deal better than supporting multiple DeferredTracer implementations.

carterkozak:
Do you have any concerns with this implementation @dansanduleac?

uschi2000 (commented):
Can we add some tests to prove that this thing is *actually* serializable?

dansanduleac:
@bavardage as @uschi2000 said, could we just copy the SerializableDeferredTracerTest from #95 ?

bavardage:
yes absolutely - I meant to copy this across but forgot. Will do this in a bit.

bavardage:
done

dansanduleac (approved):

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -49,26 +66,48 @@ public DeferredTracer(String operation) {
...
+     * If no operation is specified, will attempt to use the parent span's operation name.
+     */
+    public DeferredTracer(Optional<String> operation) {
+        Optional<Trace> maybeTrace = Tracer.copyTrace();
+        if (maybeTrace.isPresent()) {
+            Trace trace = maybeTrace.get();
+            this.traceId = trace.getTraceId();
+            this.isObservable = trace.isObservable();
+            this.parentSpanId = trace.top().map(OpenSpan::getSpanId).orElse(null);
+            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(DEFAULT_OPERATION));
```
pkoenig10, line 80: Using the previous operation when a caller does not provide one is a behavior break and seems incorrect. This results in confusing repeated spans for work that is unrelated to the operation name.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -49,26 +66,48 @@ public DeferredTracer(String operation) {
...
-        if (!trace.isPresent()) {
+        if (traceId == null) {
             return inner.call();
         }
+
         Optional<Trace> originalTrace = Tracer.copyTrace();
-        Tracer.setTrace(trace.get());
-        operation.ifPresent(Tracer::startSpan);
+
+        Tracer.setTrace(new Trace(isObservable, traceId));
```
pkoenig10, line 100: By creating a new trace here we lose the callers span stack. While the span tree derived from trace logs is correct because we set the `parentSpanId`, the span stack available at runtime is missing parent spans.

pkoenig10 (commented):
I've noticed a couple breaking behavior changes as a result of this PR that have resulted in incorrect and confusing trace logs.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
bavardage: yes, this was a known change - I'm curious what you're doing with the span stack at runtime

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
bavardage: we could make a break and require an operation name?

I'm not sure how this is very different to what was done before?

pkoenig10: Previously the deferred tracers would not create a new span if the caller did not provide an operation. This is still valid because the existing trace should already have a span.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
pkoenig10: I think this breaks an assumption that the current span stack is present in the trace. It seems dangerous and confusing for the span stack to not have the full trace information.

For example, after this PR it can be valid to have a trace with an empty span stack since this trace might actually have spans that have been dropped by the deferred tracer.

Issue #103, opened 2019-03-25
Log consumers should be able to differentiate sampled from unsampled traces


carterkozak:
## What happened?
Log lines inside a traced operation contain a `traceId` [MDC](https://www.slf4j.org/api/org/slf4j/MDC.html) key. Tracing output may be loaded by ID in our logging pipeline, however most traceIds are not sampled, and as such may not be viewed. It's difficult to tell which traceIds one can search for.

## What did you want to happen?
It would be helpful if we set a second MDC key-value pair to differentiate sampled and unsampled traces.

### Proposal

Tracer should set `sampled`:`true` when setting the `traceId` key for a sampled request.

MDC key name: `sampled`, `traceSampled`, better ideas? `sampled` is simple and easy to remember.

Should we set `sampled`:`false` for traces that are not sampled? This would allow us to differentiate versions which have not yet upgraded, however it would also slightly increase the overhead of every log line. While it may be helpful today, I don't think it's necessary long term.

carterkozak:
@dsd987 Mind taking a look at this?

carterkozak:
After some thought, I think we should use key `traceSampled` in order to make collisions with existing parameter names less likely, and provide more context describing the value.

carterkozak:
Update: Caught up with @ellisjoe who is looking into what it would take to expose this information using indexed traces without the need for additional data attached to service events.
Potential upsides:
* Less data logged per event
* Could work with request logs as well as service logs
* May give us direct UI links to trace info from service/request logs

raiju:
This PR should unblock it for Jetty filters: https://github.com/palantir/tracing-java/pull/137

PR #104, opened 2019-03-26
Adds key traceSampled to MDC when set trace is observable

dsd987:
## Before this PR
Log consumers were unable to tell which traces were sampled (observable: true) vs which traces were not sampled (observable: false). 

## After this PR
An MDC key traceSampled will now appear with the value "1" when the set trace has been sampled (observable: true). If the set trace is not sampled (observable: false), the key will not appear in the MDC.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -331,8 +331,18 @@ public final class Tracer {
     static void setTrace(Trace trace) {
         currentTrace.set(trace);
 
-        // Give SLF4J appenders access to the trace id
+        // Give SLF4J appenders access to the trace id and if trace is being sampled
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
+        setTraceSampledMdcIfObservable(trace.isObservable());
+    }
+
+    private static void setTraceSampledMdcIfObservable(boolean isObservable) {
+        if (isObservable) {
+            MDC.put(Tracers.TRACE_SAMPLED_KEY, "1");
+        } else {
+            // To ensure MDC state is cleared when trace is not observable
+            MDC.remove(Tracers.TRACE_SAMPLED_KEY);
+        }
     }
 
     private static Trace getOrCreateCurrentTrace() {
@@ -347,5 +357,6 @@ public final class Tracer {
     private static void clearCurrentTrace() {
         currentTrace.remove();
         MDC.remove(Tracers.TRACE_ID_KEY);
+        MDC.remove(Tracers.TRACE_SAMPLED_KEY);
     }
 }
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -331,13 +331,14 @@ public final class Tracer {
     static void setTrace(Trace trace) {
         currentTrace.set(trace);
 
-        // Give SLF4J appenders access to the trace id and if trace is being sampled
+        // Give log appenders access to the trace id and whether the trace is being sampled
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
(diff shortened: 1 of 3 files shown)


dsd987:
Proposed for #103

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -331,8 +331,18 @@ public static boolean isTraceObservable() {
     static void setTrace(Trace trace) {
         currentTrace.set(trace);
 
-        // Give SLF4J appenders access to the trace id
+        // Give SLF4J appenders access to the trace id and if trace is being sampled
```
carterkozak: > SLF4J appenders

No such thing!

> and if trace is being sampled

Maybe `s/if/whether the`, the first couple reads through this comment I thought there was a piece missing from the end.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -331,8 +331,18 @@ public static boolean isTraceObservable() {
...
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
+        setTraceSampledMdcIfObservable(trace.isObservable());
+    }
+
+    private static void setTraceSampledMdcIfObservable(boolean isObservable) {
```
carterkozak: nit: I don't like `is` in boolean names because it doesn't tell us anything that the type doesn't already make clear.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -331,8 +331,18 @@ public static boolean isTraceObservable() {
...
+        if (isObservable) {
+            MDC.put(Tracers.TRACE_SAMPLED_KEY, "1");
```
carterkozak, line 342: Might add a comment describing why we're using `1` instead of `true`.

tracing/src/test/java/com/palantir/tracing/TracerTest.java
```diff
@@ -190,6 +190,22 @@ public void testSetTraceSetsCurrentTraceAndMdcTraceIdKey() throws Exception {
...
+        assertThat(MDC.get(Tracers.TRACE_SAMPLED_KEY)).isNull();
+    }
+
+    @Test
+    public void testSetTraceMissingMdcTraceSampledKeyWhenNotObserved() {
+        Tracer.setTrace(new Trace(false, "notObservedTraceId"));
+        assertThat(MDC.get(Tracers.TRACE_SAMPLED_KEY)).isNull();
+        assertThat(Tracer.completeSpan()).isEmpty();
+        assertThat(MDC.get(Tracers.TRACE_SAMPLED_KEY)).isNull();
+    }
```
carterkozak, line 207: Nice tests

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -26,6 +26,7 @@
 public final class Tracers {
     /** The key under which trace ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
     public static final String TRACE_ID_KEY = "traceId";
+    public static final String TRACE_SAMPLED_KEY = "traceSampled";
```
carterkozak, line 29: We should document this key similarly to `TRACE_ID_KEY` and describe possible values.

carterkozak (commented):
Looks good, commented a few nits.

carterkozak:
@uschi2000 I'm curious of your thoughts on this.

dsd987:
Any thoughts on having the value of "0" when the trace is not sampled vs just not having the key? Would be storing more but would be more consistent with the use of the http header.

carterkozak:
On hold pending https://github.com/palantir/tracing-java/issues/103#issuecomment-476680077

carterkozak:
Closing in favor of #144

PR #106, opened 2019-03-27
[fix] Only create span in DeferredTracer if one is provided

pkoenig10:
## Before this PR
Before #97, if a caller created a `DeferredTracer` without providing an operation, no span would be created. After #97, if a caller creates a `DeferredTracer`without providing an operation, a span will be created with the span operation as the current span. This is confusing because it results in ancestries of spans having the same operation name, even though they very likely do different work.

See https://github.com/palantir/tracing-java/pull/97#discussion_r269440151.

## After this PR
The behavior of `DeferredTracer` is reverted back to what is was before #97. If a `DeferredTracer` is created without an operation, no span will be created.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -48,8 +48,6 @@ public final class DeferredTracer implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    private static final String DEFAULT_OPERATION = "deferred";
-
     @Nullable
     private final String traceId;
     private final boolean isObservable;
@@ -68,7 +66,6 @@ public final class DeferredTracer implements Serializable {
 
     /**
      * Create a new deferred tracer, optionally specifying an operation.
-     * If no operation is specified, will attempt to use the parent span's operation name.
      */
     public DeferredTracer(Optional<String> operation) {
         Optional<Trace> maybeTrace = Tracer.copyTrace();
@@ -77,7 +74,7 @@ public final class DeferredTracer implements Serializable {
             this.traceId = trace.getTraceId();
             this.isObservable = trace.isObservable();
             this.parentSpanId = trace.top().map(OpenSpan::getSpanId).orElse(null);
-            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(DEFAULT_OPERATION));
+            this.operation = operation.orElse(null);
         } else {
             this.traceId = null;
             this.isObservable = false;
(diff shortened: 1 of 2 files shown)


pkoenig10:
I'm not sure how to reconcile the changes from #97 with these failing tests. It seems like the changes in #97 allowed the tests to pass but did not maintain the intended behavior.

For example, in `testWrapRunnable_traceStateIsCapturedAtConstructionTime` we verify that we can complete a span that was started outside the deferred trace and that this completed span has the correct name. Before #97, this test completed the same span instance that was created outside the deferred trace. After #97 it is actually completing a different span instance that happens to have the same name.

An modified version of this test that would have passed before #97 but now fails can be seen here:
- Passing test on top of 2.8.0: https://github.com/pkoenig10/tracing-java/commit/fca8fa5
- Failing test on develop: https://github.com/pkoenig10/tracing-java/commit/ddd31c0

bavardage:
I think @cakofony raised the very valid point before that it's invalid to have a trace without a span, and that's what this was trying to ensure never happened.

If your concern is with misleading span names, couple of proposals
1. we could just make the break and require specifying an operation name
2. rather than just using the same name as the current span, we could prepend "deferred from" or something... at least making it explicit that it's not just the same thing

pkoenig10:
But in this case the trace really DOES have a span, it not just present in the trace instance because we've dropped the spans but kept the parentSpanId tying them together.

Having this discrepancy between what is visible the trace instance at runtime and what is actually in the trace seems to break a basic assumption of this library. Have we decided that it is now valid for the bottom span in the stack to have a `parentSpanId`? I don't have any particular problem with this new approach, but this seems like a pretty significant change in behavior.

At the very least we should not be reusing the operation name. But requiring a span on a deferred trace isn't really necessary in many cases and just leads to more noise in the trace logs. See #89 as an example.

pkoenig10:
I definitely agree it should be invalid to have a *trace* (all spans with a particular trace id) without a span. But the question is, can it be valid to have a *trace instance* (the trace data present at runtime) without a span so long as the former requirement is met. I would answer yes.

carterkozak:
>  Have we decided that it is now valid for the bottom span in the stack to have a parentSpanId? 

Yes the bottom span is allowed to have a parentId, that's how requests received by the TraceEnrichingFilter creates new spans with info from tracing aware clients!

It sounds like the problem is that a DeferredTrace without its own span causes the first inner span that completes to clear the entire trace -- that would be bad. We can resolve this by pushing a dummy span onto the stack, it doesn't need any real data, it only serves to avoid triggering what the thread considers completion of a trace (when a root span is completed, and thread state is cleared)

pkoenig10:
Okay that makes sense. So my only request is that we don't reuse the parent operation in the deferred tracer. I'll open a new PR.

pkoenig10:
Moved to #110

PR #110, opened 2019-03-27
[fix] Don't use parent span operation in deferred traces

pkoenig10:
## Before this PR
When creating a deferred trace without specifying an operation, the parent span operation is used. This is confusing because:
1. It creates duplicate spans with the same operation name
2. The parent span operation name is likely not a good description of what the deferred trace task is doing.

## After this PR
When creating a deferred trace without specifying an operation, the default deferred span operation is used. The constructors that allow callers to not specify an operation have been deprecated a signal to clients that they should provide an operation for better trace logs.

### DIFF
--- a/tracing/src/test/java/com/palantir/tracing/TracersTest.java
+++ b/tracing/src/test/java/com/palantir/tracing/TracersTest.java
@@ -62,15 +62,15 @@ public final class TracersTest {
                 Tracers.wrap(Executors.newSingleThreadExecutor());
 
         // Empty trace
-        wrappedService.submit(traceExpectingCallableWithSingleSpan("deferred")).get();
-        wrappedService.submit(traceExpectingCallableWithSingleSpan("deferred")).get();
+        wrappedService.submit(traceExpectingCallableWithSingleSpan("DeferredTracer(unnamed operation)")).get();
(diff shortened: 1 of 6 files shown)


tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -203,7 +203,7 @@ public void testWrapCallable_callableTraceIsIsolated() throws Exception {
     public void testWrapCallable_traceStateIsCapturedAtConstructionTime() throws Exception {
         Tracer.startSpan("before-construction");
         Callable<Void> callable = Tracers.wrap(() -> {
-            assertThat(Tracer.completeSpan().get().getOperation()).isEqualTo("before-construction");
+            assertThat(Tracer.completeSpan().get().getOperation()).isEqualTo("deferred");
```
iamdanfox: When the user doesn't supply an operation name, what do you think about using `DeferredTracer(unnamed operation)`.  I hope this would be enough information for people to hunt down usages of 'DeferredTracer' in their codebase and hopefully be prompted to give them an operation name?

pkoenig10:
Closing in favor of #106

ferozco:
👍

PR #117, opened 2019-03-29
[improvement] Add AsyncTracer that tracks enqueued-time and run-time of a deferred action

robert3005:
## Before this PR
DeferredTracer only preserves traceid and last span id

## After this PR
Add AsyncTracer that preserves full span stack of parent trace

cc @ellisjoe

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
@@ -1,0 +1,67 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+import java.util.Optional;
+
+public final class AsyncTracer {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final String DEFAULT_OPERATION = "deferred";
+
+    private final Trace deferredTrace;
+    private final String operation;
+
+    public AsyncTracer() {
+        this(Optional.empty());
+    }
+
+    public AsyncTracer(String operation) {
+        this(Optional.of(operation));
+    }
+
+    /**
+     * Create a new deferred tracer, optionally specifying an operation.
+     * If no operation is specified, will attempt to use the parent span's operation name.
+     */
+    public AsyncTracer(Optional<String> operation) {
+        this.operation = operation.orElse(DEFAULT_OPERATION);
+        Tracer.startSpan(this.operation + "-enqueue");
+        deferredTrace = Tracer.copyTrace().get();
+        Tracer.fastDiscardSpan(); // span will completed in the deferred execution
+    }
+
+    /**
+     * Runs the given callable with the current trace at
+     * the time of construction of this {@link DeferredTracer}.
+     */
+    public <T, E extends Throwable> T withTrace(Tracers.ThrowingCallable<T, E> inner) throws E {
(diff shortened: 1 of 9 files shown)


carterkozak:
Why is it important that we can start a span, then complete it on multiple threads? tracing-java makes assumptions about threads that become foot-guns when applied to asynchronous usage. It might be time for a v2 API.

ellisjoe:
Yeah, we're explicitly interested in tracing async things (eg: how long tasks spend in queues or waiting for their client side limiters)

carterkozak:
Can we produce a new function for this case with fewer sharp edges? I'm worried about queueing multiple operations, and having them all complete copies of the same span. We need some way to enforce the concept of span ownership.

robert3005:
@cakofony what is the case you currently have that you're worried about? Isn't it the case that multiple threads can finish copies of the span before #97

carterkozak:
Yes, that was a bug and a foot-gun.

Perhaps this behavior should live in a separate `@Beta` tracer type which doesn't need to be serializable. It's not clear to me that DeferredTracer is the right home for this functionality. We don't want to defer a trace, we're elongating a span.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -50,13 +44,8 @@
...
     private static final String DEFAULT_OPERATION = "deferred";
 
-    @Nullable
-    private final String traceId;
-    private final boolean isObservable;
-    @Nullable
+    private final Trace deferredTrace;
     private final String operation;
-    @Nullable
-    private final String parentSpanId;
```
carterkozak: If we go with this approach it will require a `serialVersionUID` bump

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -71,48 +60,28 @@ public DeferredTracer(String operation) {
...
-            this.isObservable = trace.isObservable();
-            this.parentSpanId = trace.top().map(OpenSpan::getSpanId).orElse(null);
-            this.operation = operation.orElse(trace.top().map(OpenSpan::getOperation).orElse(DEFAULT_OPERATION));
-        } else {
-            this.traceId = null;
-            this.isObservable = false;
-            this.parentSpanId = null;
-            this.operation = null;
-        }
+        this.operation = operation.orElse(DEFAULT_OPERATION);
```
pkoenig10: Forwarding @dfox's comment about the default operation name: https://github.com/palantir/tracing-java/pull/110#discussion_r269833764

carterkozak:
Rough notes on the detached span idea so I remember later:

```java
/* Does not modify thread state, the current span is used as detached spans parent
 * if it exists, otherwise a detached span is created with a new trace that has no
 * thread state. */
DetachedSpan detached = Tracer.startDetachedSpan("operation")
```

```java
interface DetachedSpan {
  /* Executes the Runnable with tracing state applied. */
  void run(String operation, Runnable runnable)

  /* Executes the Callable with tracing state applied. */
  <T> T call(String operation, Callable<T> callable)

  /* Complete this span */
  void close();
}
```

markelliot (approved):

carterkozak:
Thanks @robert3005 👍

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -141,6 +141,14 @@ private static OpenSpan startSpanInternal(String operation, SpanType type) {
         return span;
     }
 
+    /** Discards the current span without emitting it. */
+    static void fastDiscardSpan() {
+        Trace trace = currentTrace.get();
+        if (trace != null) {
+            popCurrentSpan();
```
dansanduleac: let's throw here, we never expect it to be null

dansanduleac (approved):

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
dansanduleac: I missed this sorry!

tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+@Beta
+public final class AsyncTracer {
+
+    private static final String DEFAULT_OPERATION = "async";
+
+    private final Trace deferredTrace;
+    private final String operation;
+
+    public AsyncTracer() {
```
pkoenig10, line 51: Given that this is a new class, we should have made the operation required.

PR #118, opened 2019-03-29
Add tests for async observer executor shutdown

schlosna:
## Before this PR

There was no unit test coverage for cases where the async observer executor was shutdown before spans completed.

## After this PR

Additional test coverage for #66 & #67 for when the async observer executor is shutdown.

### DIFF
--- a/tracing/src/test/java/com/palantir/tracing/AsyncSlf4jSpanObserverTest.java
+++ b/tracing/src/test/java/com/palantir/tracing/AsyncSlf4jSpanObserverTest.java
@@ -30,6 +30,8 @@ import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.Appender;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.tracing.api.OpenSpan;
 import com.palantir.tracing.api.Span;
 import com.palantir.tracing.api.SpanType;
 import java.net.Inet4Address;
@@ -37,6 +39,7 @@ import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 import org.jmock.lib.concurrent.DeterministicScheduler;
 import org.junit.After;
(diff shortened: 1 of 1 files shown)


tracing/src/test/java/com/palantir/tracing/AsyncSlf4jSpanObserverTest.java
```diff
@@ -219,6 +222,38 @@ public void testSpanMetadataConvertsToZipkinBinaryAnnotations() {
...
+                "serviceName", Inet4Address.getLoopbackAddress(), logger, executor));
+        OpenSpan span1 = Tracer.startSpan("operation");
+        assertThat(Tracer.completeSpan()).isPresent().get()
+                .extracting("spanId", "operation")
+                .contains(span1.getSpanId(), span1.getOperation());
+
+        OpenSpan span2 = Tracer.startSpan("operation");
+        executor.shutdown();
+        assertThat(Tracer.completeSpan()).isPresent().get()
+                .extracting("spanId", "operation")
```
carterkozak, line 239: Missing a tracer.unsubscribe. If handling works as expected that shouldn’t be allowed to break anything, but may cause confusing logging depending on test order

tracing/src/test/java/com/palantir/tracing/AsyncSlf4jSpanObserverTest.java
```diff
@@ -219,6 +222,38 @@ public void testSpanMetadataConvertsToZipkinBinaryAnnotations() {
...
+        Tracer.subscribe("", AsyncSlf4jSpanObserver.of(
+                "serviceName", Inet4Address.getLoopbackAddress(), logger, executor));
+        OpenSpan span1 = Tracer.startSpan("operation");
+        assertThat(Tracer.completeSpan()).isPresent().get()
+                .extracting("spanId", "operation")
+                .contains(span1.getSpanId(), span1.getOperation());
+
+        Tracer.startSpan("operation");
+        executor.shutdown();
+        Tracer.fastCompleteSpan();
```
carterkozak, line 255: Here as well

carterkozak:
:+1:

tracing/src/test/java/com/palantir/tracing/AsyncSlf4jSpanObserverTest.java
schlosna: moved the `unsubscribe`s to `before`/`after` to ensure clean state

PR #124, opened 2019-04-01
Support for tracing asynchronous operations

carterkozak:
Added a Tracer API to create a span that is not attached to a
thread.


tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -0,0 +1,36 @@
...
+ * Span operation which is not bound to thread state, and can measure operations which
+ * themselves aren't bound to individual threads.
+ */
+public interface DetachedSpan {
+
+    /**
+     * Equivalent to {@link Tracer#startSpan(String)}, but using this {@link DetachedSpan} as the parent instead
+     * of thread state.
+     */
+    SpanToken startSpan(String operationName);
```
carterkozak: todo: `startSpan(operation, type)`

carterkozak: This API uses a Closeable token rather than `runWith(operation, runnable)` or `callWith(operation, callable)` because it's more versatile in some situations when the operation we wish to trace is not itself a standard java runnable/callable, though we lose some comile-time safety that the token is not passed to another thread.

tracing/src/main/java/com/palantir/tracing/SpanToken.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.palantir.tracing;
+
+import java.io.Closeable;
+
+/**
+ * Closeable marker around a tracing span operation. This object should be used in a try/with block.
+ */
+public interface SpanToken extends Closeable {
```
carterkozak: SpanToken vs TraceToken? Closing this represents closing a span, however it also cleans up the entire trace state. Thoughts on a better name? This class is unnecessary if we use the `Runnable`/`Callable` approach.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -123,6 +125,90 @@ public static OpenSpan startSpan(String operation) {
         return startSpanInternal(operation, SpanType.LOCAL);
     }
 
+    /**
+     * Like {@link #startSpan(String, SpanType)}, but does not set or modify tracing thread state.
+     */
+    public static DetachedSpan startDetachedSpan(String operation, SpanType type) {
```
carterkozak: This might be clearer if we call it `startAsyncSpan` instead.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -123,6 +125,90 @@ public static OpenSpan startSpan(String operation) {
...
+                    ? maybeCurrentTrace.isObservable() : sampler.sample();
+            openSpan = OpenSpan.builder()
+                    .parentSpanId(getParentSpanId(maybeCurrentTrace))
+                    .spanId(Tracers.randomId())
+                    .operation(operation)
+                    .type(type)
+                    .build();
+        }
+
+        /* n.b. This implementation does not preserve and replace spans that exist when the operation is started. */
```
carterkozak: This should be implemented if we decide to use this approach.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
carterkozak: This method name should make it clear that we're setting state on the current thread. `startSpanWithThreadState`?

tracing/src/main/java/com/palantir/tracing/SpanToken.java
carterkozak: I think Span is most correct here, these may be nested spans which are part of the same trace.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: However it's not necessarily asynchronous, it's no different from any other span except that it is not connected to any thread-local state.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -0,0 +1,46 @@
...
+
+package com.palantir.tracing;
+
+import com.palantir.tracing.api.SpanType;
+
+/**
+ * Span operation which is not bound to thread state, and can measure operations which
+ * themselves aren't bound to individual threads.
+ */
+public interface DetachedSpan {
```
carterkozak, line 24: We may want an API to create a child `DetachedSpan` from this one without modifying state.
This could be accomplished indirectly using the following, however it requires an intermediate operation. I am hesitant to add an API to apply our span to a thread without also applying a new span.
```java
try (SpanToken ignored = detached.startSpanOnCurrentThread("operation") {
    return Tracer.startDetachedSpan("nextOperation");
}
```

tracing/src/main/java/com/palantir/tracing/SpanToken.java
markelliot: why is a `SpanToken` required? (why can't we use `OpenSpan`?)

markelliot:
perhaps an alternative here would be exposing `Trace` + having `Trace#createSpan` exposed, and we could add a method `copyOrCreateTrace` which returns a new `Trace` captured from the thread state, but is otherwise 'detached' and still mutable in the way that's required by AsyncTrace

tracing/src/main/java/com/palantir/tracing/SpanToken.java
carterkozak: OpenSpan is a data object, I want to expose an object whose only job is to complete an operation, with the ability to add validation that we have done so correctly.
We could use the static `Tracer.completeSpan()` method instead for this case because we do need to clean up thread state, however I've fixed too many tracing thread state leaks to be comfortable with that approach. Using the proposed API we can validate that we are closing the same span that we opened, which is not something we can do using the existing API.

carterkozak:
I'd prefer not to expose the `Trace` or `OpenSpan` objects directly because API consumers shouldn't need access to that internal data to instrument their code. I realize these are already exposed to some extent, but we can avoid further api/implementation coupling while providing both easier and safer utilities.

tracing/src/main/java/com/palantir/tracing/SpanToken.java
markelliot: Why do we even need `SpanToken`? You can just return a `Closeable` anywhere you return `SpanToken`?

carterkozak: Closeable can throw IOException

markelliot:
`DetachedSpan` seems like as odd a thing to expose as `Trace` or `OpenSpan`. I guess the thing I'm poking at is that in cases where we want traces to be detached from threads, we'd logically be implementing exactly the things that we attach to threads. I'd rather we expose the basics than invent other concepts More generally, what seems appropriate for API surface here seems to follow a pretty arbitrary set of rules -- the initial intent was something extremely limited, and the drift that's grown with the library (from having basically three methods) now makes it hard to hold a reasonable bar.

carterkozak:
Agreed that we're in an odd place, and there's not a lot we can do without either breaking API or creating sprawl. The `Trace` object doesn't work well across threads, it's meant to represent a single threads state -- at it's core it's a tuple of `traceId`, `sampled`, `stack<open-span>`. The stack maps directly to the call stack, copying this stack across threads destroys this mapping and makes it easy to shoot ourselves in the foot.

If we wanted to focus on safety, the `Trace` object would take the form `{traceId, sampled, currentSpan}`, and tracing operations would be encapsulated in a runnable e.g. `Tracer.startSpan("operation", () -> myTracedOperation);` where the Trace stack is replaced by span invocations on the application stack. In this model, the proposed `DefaultDetachedSpan` would be unnecessary, and the Trace object could make sense. Unfortunately we have taken advantage of too many sharp edges of the current Trace object (e.g. AsyncTracer) to make this change without a major rev, which is likely more trouble than it's worth at the moment.

blab: I don't have a great solution.

ellisjoe:
Hey, just jumping in on this now since I'm going to need something like this for [dialogue](https://github.com/palantir/dialogue). Has any more thought gone into this issue, or are the comments here up to date and we just need to come to an agreement on some proposal?

carterkozak:
Hey @ellisjoe, a month or so ago I caught up with @markelliot and we discussed a couple models. Mark was planning to put together a sample proposal/implementation that included a concept of span ownership, but he may have been pulled into other things. I'd be happy to chat through possibilities.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Equivalent to {@link Tracer#startSpan(String, SpanType)}, but using this {@link DetachedSpan}
+     * as the parent instead of thread state.
+     */
+    @MustBeClosed
+    SpanToken attach(String operationName, SpanType type);
```
ellisjoe: What's the difference between calling `detach(operation)` and then `complete()` vs `attach(operation)` and then `close()`? is it just that calling attach also overrides my current thread local state?

carterkozak: attach applies a _new_ span to the current thread, from `attach()` to `close()` This can be done multiple times to create sibling spans with the `DetachedSpan` as a parent.

`DetachedSpan.complete`, is used to complete the detached span, which is not associated with a thread. There is danger that complete is never called, and we fail to associate the child spans (from `attach`) with the original parent of this detached span, but I don't think there are good options to avoid that.

This model means that you cannot ever detach or attach an existing span that is already bound to a thread, each time that we attach or detach we mark a new operation. I've found that otherwise, it can become difficult to trace where our tracing spans have actually come from.

ellisjoe: Oh I see, so you'd start your `DetachedSpan`, then you could `attach()` it once your async task starts running, and then go back to using the normal `Tracer.startSpan()` methods throughout.

Think the thing that seems a bit funky is that you need to close the `SpanToken` rather than just using `Tracer.completeSpan()`.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    default DetachedSpan detach(String operation) {
+        return detach(operation, SpanType.LOCAL);
+    }
+
+    /**
+     * Completes this span. After complete is invoked, other methods are not expected to produce spans, but
+     * they must not throw either in order to avoid confusing failures.
+     */
+    void complete();
```
ellisjoe, line 97: any reason for this class not being `Closeable` instead?

carterkozak: Nope, that would be cleaner 👍

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
carterkozak: That is correct, the reason for the different type is so we can reset the original tracing data on the attached thread. We could ignore that, and apply state directly, but it would be confusing to api consumers whether or not the attach creates a new tracing span, and how many they are responsible for completing.

ellisjoe: I think this may have been more obvious to me if the `attach()` method was on `Tracers` rather than here. In my head, at least at the moment, `Tracers` deals with all the threading business, and so I could imagine wanting to call `Tracers.attach(detachedSpan)` which would then give me a `SpanToken` to close and reset the thread's trace. And then on `DetachedSpan` you just would just have `newSpan()` which gives you a new `DetachedSpan` parented by the current one. That way `DetachedSpan` has nothing to do with threading.

carterkozak: 👍 That sounds good to me.

ellisjoe: As a point of reference, that also seems to be what these guys are doing as well: https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/Tracer.java#L140

carterkozak:
Holding off fixing this up until #180 has gone in to unblock other work

carterkozak:
Rebased, have to catch a flight shortly, will knock out the rest when I can.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
carterkozak: `Tracers.attach(detachedSpan)` would also need an operation name param. It would also require us to pass some additional data between DetachedSpan and Tracer, which isn't currently possible. Right now, the DetachedSpan interface doesn't leak whether or not it is sampled, but that _is_ required in order to attach to a thread.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -19,13 +19,33 @@
...
+    static DetachedSpan start(String operation, SpanType type) {
+        return Tracer.detachInternal(operation, type);
+    }
+
+    /**
+     * Opens a new {@link SpanType#LOCAL LOCAL} detached span for this thread's call trace,
+     * labeled with the provided operation.
+     */
+    @CheckReturnValue
+    static DetachedSpan start(String operation) {
```
carterkozak: @ellisjoe @ferozco I've renamed these from `detach` to `start`, otherwise method names conflict with non-static methods. I think `DetachedSpan.start` conveys the detached nature of the span as well as `Tracer.detach`.

carterkozak:
I've updated the API, there's still some implementation code that I need to move around. I don't like having the DetachedSpan implementations living in Tracer, but I can clean that up once we're happy with the ergonomics.

markelliot:
I think the challenge here is just how much duplication now exists. It seems likely we’re going to run into problems keeping parallel implementations of attached/detached spans+traces. 

If the present APIs don’t work, I think we’d be better off making a next major rev of the library rather than adding substantial cruft.

ferozco:
@markelliot While I'm sympathetic to your concern, I think what we've arrived to here is a small expansion of the API and unblocks high value work that in my mind outweighs the additional maintenance burden. I do agree that we should revisit the API and clean it up but I'd like to do that once we've had some time get experience with the new APIs within CJR or Dialogue.

iamdanfox:
:-1: Just wanna put a little blocker on merging this because I tried to use it in a little demo and pretty surprised by the behaviour... will post the output in a sec

carterkozak:
oh geez. One character.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -172,7 +172,7 @@ static DetachedSpan detachInternal(String operation, SpanType type) {
 
     private static Optional<String> getParentSpanId(@Nullable Trace trace) {
         if (trace != null) {
-            Optional<OpenSpan> maybeOpenSpan = trace.pop();
+            Optional<OpenSpan> maybeOpenSpan = trace.top();
```
markelliot, line 208: are there tests to catch this sort of thing?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -154,6 +157,105 @@ public static void fastStartSpan(String operation) {
...
+                    .build();
+        }
+
+        @Override
+        public SpanToken attach(String operationName, SpanType type) {
+            warnIfCompleted("startSpanOnCurrentThread");
+            Trace maybeCurrentTrace = currentTrace.get();
+            Trace trace = Trace.of(sampled, traceId);
+            setTrace(trace);
+            Tracer.fastStartSpan(operationName, type);
```
ferozco: You also will want to explicitly pass the spanId of `openSpan` as the parentId of the new span

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: yep

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -154,6 +157,139 @@ public static void fastStartSpan(String operation) {
...
+            this.openSpan = OpenSpan.builder()
+                    .parentSpanId(parentSpanId)
+                    .spanId(Tracers.randomId())
+                    .operation(operation)
+                    .type(type)
+                    .build();
+        }
+
+        @Override
+        public SpanToken attach(String operationName, SpanType type) {
```
ferozco: Do you think it would be more ergonomic to "complete" the detached span when attaching it? I think this would encourage using the detachedSpan to indicate the amount of time spent between workers. i.e. some producers putting detachedSpans on queue somewhere and then having consumers pop them off and attach them.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -154,6 +157,139 @@ public static void fastStartSpan(String operation) {
...
+        if (trace != null) {
+            Optional<OpenSpan> maybeOpenSpan = trace.top();
+            if (maybeOpenSpan.isPresent()) {
+                return Optional.of(maybeOpenSpan.get().getSpanId());
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static final class SampledDetachedSpan implements DetachedSpan {
```
ferozco, line 216: Could we convert DetachedSpan into an abstract class, and inline Smapled/UnsampledDetachedSpan there? It would make the layout consistent with how Trace is currently implemented

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -154,6 +157,139 @@ public static void fastStartSpan(String operation) {
...
+        public SpanToken attach(String operationName, SpanType type) {
+            warnIfCompleted("startSpanOnCurrentThread");
+            Trace maybeCurrentTrace = currentTrace.get();
+            setTrace(Trace.of(true, traceId));
+            Tracer.fastStartSpan(operationName, openSpan.getSpanId(), type);
+            return TraceRestoringSpanToken.of(maybeCurrentTrace);
+        }
+
+        @Override
+        public DetachedSpan detach(String operation, SpanType type) {
```
ferozco: What use case do you envision for this?

carterkozak: Consider a webserver: When a request is received on a non-blocking thread, I'd like to start a detached span representing the entire request. Later, before processing, the request is enqueued to a larger thread pool where we allow work -- I'd like the ability to track time spent enqueued, despite not having attached the span to thread.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: It would absolutely be more ergonomic, however it would not solve a set of the cases that this model is meant to address.
A standard attached span can have multiple children, I think it's similarly helpful to support multiple children using detached spans.

Given the DetachedSpan API, we can build ergonomic helpers which do precisely what you have suggested.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: I could go either way on this. the DetachedSpan interface is meant to be relatively straightforward to read, and detached from implementation details because it's an API that will be consumed by developers using this library. Trace is an implementation detail, and isn't really consumed outside of this library.

If we moved these to the DetachedSpan class, we would also need to implement a package private accessor for the raw value of the Tracer trace thread-local, which could be dangerous if folks try to get clever.

iamdanfox:
Unblocking :+1: I think the next steps are to publish an RC from here and then make sure it looks sensible in conjure-java-runtime.

iamdanfox:
Dan and I are trying this out on conjure-java-runtime's https://github.com/palantir/conjure-java-runtime/compare/ds/use-new-tracing?expand=1, findings so far:

- ~we want to use CloseableSpan instead of the `Tracer#startSpan` to get the X-B3-SpanId etc headers onto the wire, but don't have an easy way to get the current header.~ released as 3.2.0-rc3
- maybe we should use the java `assert` keyword to make sure people wire up their instrumentation correctly in tests?

ferozco:
👍

PR #125, opened 2019-04-03
Delete redundant file

iamdanfox:
We moved to gradle-consistent-versions, this old nebula lock file is never read anymore.


carterkozak:
👍

PR #128, opened 2019-04-03
Revert okhttp bump, don't auto upgrade external libs

dansanduleac:
## Before this PR

#99 has bumped us to okhttp 3.14.0 which seems to have brought in a bunch of behaviour breaks.

## After this PR

We're reverting #99.
Also, turning off dependency-upgrader for external libraries, keeping it on only for compileOnly / test dependencies.


carterkozak:
Are we tracking breaks in okhttp 3.14 somewhere?

carterkozak:
👍 tyvm

dansanduleac:
Yes, we've seen some internally - admittedly it could be due to how we use it, but in any case a low-level library like this one shouldn't be the one that transitively bumps such dependencies.

versions.props
```diff
@@ -1,18 +1,21 @@
+# dependency-upgrader:OFF
```
iamdanfox, line 1: So the one downside of this is that we don't find out about deprecations early...

Can we include a comment saying "please only bump these if you need new features, as this is a very low level library and it will transitively upgrade the entire world"

iamdanfox (approved):

PR #135, opened 2019-04-10
[experiment] Try building against java 12

iamdanfox:
Internally, we're pushing all our services to run on Java11 _only_ as it's a nice LTS release.

For _libraries_ however, we want to keep testing against Java8 to make sure we don't break backwards compatibility. Given we have to keep the java 8 CI, I think we might as well start testing against Java 12, so hopefully we can catch any tooling gotchas early.

cc @robert3005

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,25 +32,9 @@ buildscript {
     }
 }
 
-apply plugin: 'com.palantir.git-version'
-
+apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.consistent-versions'
-
-// expanded version of 'com.palantir.baseline' because errorprone doesn't work on Java12 yet
-apply plugin: 'com.palantir.baseline-config'
-apply plugin: 'com.palantir.baseline-circleci'
-allprojects {
-    apply plugin: 'com.palantir.baseline-checkstyle'
-    apply plugin: 'com.palantir.baseline-scalastyle'
-    apply plugin: 'com.palantir.baseline-eclipse'
-    apply plugin: 'com.palantir.baseline-idea'
-    // apply plugin: 'com.palantir.baseline-error-prone'
-    // apply plugin: 'com.palantir.baseline-versions'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -33,9 +33,25 @@ buildscript {
...
 
+// expanded version of 'com.palantir.baseline' because errorprone doesn't work on Java12 yet
+apply plugin: 'com.palantir.baseline-config'
+apply plugin: 'com.palantir.baseline-circleci'
+allprojects {
+    apply plugin: 'com.palantir.baseline-checkstyle'
+    apply plugin: 'com.palantir.baseline-scalastyle'
+    apply plugin: 'com.palantir.baseline-eclipse'
+    apply plugin: 'com.palantir.baseline-idea'
+    // apply plugin: 'com.palantir.baseline-error-prone'
```
iamdanfox: Seems like it's a known issue that error-prone 2.3.3 doesn't work with Java 12: https://github.com/google/error-prone/issues/1106

build.gradle
```diff
@@ -67,5 +67,6 @@ subprojects {
 
     tasks.withType(JavaCompile) {
         options.compilerArgs += ['-Xlint:deprecation', '-Werror']
+        options.errorprone.errorproneArgs += '-Xep:Finally:OFF'
```
iamdanfox, line 72: I actually just tried this on a much larger internal project (SMM) and errorprone passed nicely when I turned off this single rule!

gm2211:
Adding to this, another large internal project (ICS server) has source `11` and compiling with `jdk11` is super slow, but it's really fast with `jdk12` (think 3+ minutes in `jdk11` and < 20 seconds in `jdk12`). Can use `jdk12` fine in intellij, but it fails in gradle with:
```
     Please report this at https://github.com/google/error-prone/issues/new and include the following:

     error-prone version: 2.3.3
     BugPattern: Finally
     Stack Trace:
     java.lang.NoSuchFieldError: label
```

PR #137, opened 2019-04-12
[improvement] Add sampled-state to filters

raiju:
## Before this PR

The jetty filter doesn't add the sampled state of the trace the request context.

## After this PR

The jetty filter adds the sampled state to the request context, enabling e.g. logging to mark sampled traces for later inspection/filtering.

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
@@ -25,6 +25,7 @@ import com.palantir.tracing.api.SpanType;
 import com.palantir.tracing.api.TraceHttpHeaders;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
+import io.undertow.util.AttachmentKey;
 import io.undertow.util.HeaderMap;
 import io.undertow.util.HttpString;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
```diff
@@ -62,8 +62,11 @@ public TracedOperationHandler(HttpHandler delegate, String operation) {
     @Override
     public void handleRequest(HttpServerExchange exchange) throws Exception {
         String traceId = initializeTrace(exchange);
+        String sampled = Tracer.isTraceObservable() ? "1" : "0";
+
         // Populate response before calling delegate since delegate might commit the response.
         exchange.getResponseHeaders().put(TRACE_ID, traceId);
+        exchange.getResponseHeaders().put(IS_SAMPLED, sampled);
```
carterkozak: Can we create an attachment for this instead of using the response headers?

raiju: Done. Is this how attachments should be used?

carterkozak: Yep, thanks!

carterkozak:
lgtm 👍

PR #142, opened 2019-04-16
[improvement] Add option to not create run spans in AsyncTracer

robert3005:
## Before this PR
AsyncTracer would always create two spans "-enqeueu" and "-run"

## After this PR
Users can indicate whether they want the additional "-run" span

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
@@ -47,6 +47,7 @@ public final class AsyncTracer {
 
     private final Trace deferredTrace;
     private final String operation;
+    private final boolean createRunSpan;
 
     public AsyncTracer() {
         this(Optional.empty());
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
```diff
@@ -56,12 +57,21 @@ public AsyncTracer(String operation) {
         this(Optional.of(operation));
     }
 
+    public AsyncTracer(String operation, boolean createRunSpan) {
+        this(Optional.of(operation), createRunSpan);
```
carterkozak, line 61: this ctor should house the full implementation, optional ctor just passes `operation.orElse(DEFAULT_OPERATION)` for backcompat

carterkozak (commented):
This is more confusing than #124

tracing/src/main/java/com/palantir/tracing/AsyncTracer.java
```diff
@@ -76,12 +86,17 @@ public AsyncTracer(Optional<String> operation) {
         Tracer.setTrace(deferredTrace);
         // Finish the enqueue span
         Tracer.fastCompleteSpan();
-        Tracer.startSpan(operation + "-run");
```
carterkozak, line 79: bug: `-run` doesn't always have the same traceId as `-enqueue`

robert3005:
You're right - I forgot the conversation we had about this previously

PR #143, opened 2019-04-16
[improvement] Use java Instant to set startTimeMicroSeconds

diogoholanda:
## Before this PR
`startTimeMicroSeconds` is set using `System.currentTimeMillis()`, which only gives millisecond granularity

## After this PR
`startTimeMicroSeconds` is set using `Instant.now()`. On Java 8 it still only gives millisecond granularity, but it gives microsecond granularity in java 11.

## Concerns
I am assuming it is fair game to use Java 8 classes in this repo

### DIFF
--- a/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
@@ -16,6 +16,7 @@
 
 package com.palantir.tracing.api;
 
+import java.time.Instant;
 import java.util.Optional;
 import org.immutables.value.Value;
 
@@ -71,8 +72,13 @@ public abstract class OpenSpan {
     public static Builder builder() {
         return new Builder()
                 // TODO(rfink): Use direct access to system microseconds when moving to Java8 / Java9
-                .startTimeMicroSeconds(System.currentTimeMillis() * 1000)
+                .startTimeMicroSeconds(getNowInMicroSeconds())
                 .startClockNanoSeconds(System.nanoTime());
+    }
+
+    private static long getNowInMicroSeconds() {
+        Instant now = Instant.now();
+        return (1000000 * now.getEpochSecond()) + (now.getNano() / 1000);
     }
 
     public static class Builder extends ImmutableOpenSpan.Builder {}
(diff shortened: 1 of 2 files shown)


tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
```diff
@@ -71,9 +72,14 @@
...
                 // TODO(rfink): Use direct access to system microseconds when moving to Java8 / Java9
-                .startTimeMicroSeconds(System.currentTimeMillis() * 1000)
+                .startTimeMicroSeconds(getNowInMicroSeconds())
                 .startClockNanoSeconds(System.nanoTime());
     }
 
+    private static long getNowInMicroSeconds() {
+        Instant now = Instant.now();
+        return (1000000 * now.getEpochSecond()) + (now.getNano() / 1000);
+    }
```
carterkozak, line 83: We run this hundreds to thousands of times per request, regardless of whether our trace is sampled.
Does this have a measurable performance impact?

diogoholanda: I have not tested the performance impact. Do you have a suggested way of doing that?

Also, as a separate thing issue, should we stop calculating those timestamps for unsampled traces?

carterkozak: Looks like a 20-25% regression in clock time on java 8, and 50% regression on java 12. I think that's a reasonable cost for more accurate values. If we see problems we can optimize by not calculating times for unsampled traces.

iamdanfox:
If we _really_ want to be able to order spans within a millisecond, we could actually just record the starting System.nanoTime() to the Span class, then actually figure out the true clock time inside the SpanObservers (as these happen on other threads, and we would have decided whether the trace is sampled by then?)

carterkozak:
I'd prefer to use the jvm utility meant for this, we'll hit all kinds of fun bugs if we try to re-build it ourselves ;-)

iamdanfox:
Ok LGTM :+1: most requests are still dominated by other upstream network calls, so taking a small perf hit here is probably worth it if the added precision helps debug and eliminate egregious calls

diogoholanda:
@iamdanfox I don't have merge permissions. Could you merge this one?

PR #144, opened 2019-04-17
[improvement] Adds key traceSampled to MDC when set trace is observable

raiju:
## Before this PR
Log consumers were unable to tell which traces were sampled (observable: true) vs which traces were not sampled (observable: false).

## After this PR
An MDC key traceSampled will now appear with the value "1" when the set trace has been sampled (observable: true). If the set trace is not sampled (observable: false), the key will not appear in the MDC.

Note: this replaces #104


carterkozak:
👍

PR #145, opened 2019-04-18
[Improvement] Bump TraceEnrichingFilter priority

raiju:
## Before this PR

The `TraceEnrichingFilter` filter can be after filters consuming tracing state.

## After this PR

The filter is now installed with a higher priority.


carterkozak:
👍

PR #147, opened 2019-04-24
[proposal] Allow creating deferred traces without spans

pkoenig10:
This PR probably should not be merged in the current form but instead should be replaced by a more appropriate API. The purpose of this PR is to facilitate a discussion about how best to support this behavior.

At the very least, this use case should be taken into consideration in https://github.com/palantir/tracing-java/pull/124.

## Before this PR
It was not possible to create a wrapped runnable/callable/executor that did not create a new span for the task being run. This behavior used to be optional, but it was made mandatory in #97. While ideally, we want every deferred task to have a span, this can cause ergonomic issues when chaining asynchronous tasks.

For example, when chaining multiple futures:
```
FluentFuture.from(future)
    .transform(result -> handleResult1(...), wrappedExecutor)
    .transform(result -> handleResult2(...), wrappedExecutor)
```

Each executor tasks submits the callback inside it's own task. This results in nested executor spans as deep as the callback chain. For the above example, the spans would look like:
```
parentSpan
|-- executor
    |-- handleResult1
    |-- executor
         |-- handleResult2
```

In this case, it would be preferable to have an executor that did not wrap tasks in a new span.

## After this PR
`DeferredTracer` allows running tasks that will not create a new containing span.

For the above example, the spans would now look like:
```
parentSpan
|-- handleResult1
|-- handleResult2
```

### DIFF
--- a/tracing-jaxrs/src/test/java/com/palantir/tracing/jaxrs/JaxRsTracersTest.java
+++ b/tracing-jaxrs/src/test/java/com/palantir/tracing/jaxrs/JaxRsTracersTest.java
@@ -19,6 +19,7 @@ package com.palantir.tracing.jaxrs;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.palantir.tracing.Tracer;
+import com.palantir.tracing.api.Span;
 import java.io.ByteArrayOutputStream;
 import javax.ws.rs.core.StreamingOutput;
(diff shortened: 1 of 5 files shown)


iamdanfox:
Closing out because we have DetachedSpan now https://github.com/palantir/tracing-java/pull/124

Issue #32, opened 2018-11-01
Support undecided observability state


carterkozak:
The default sampler should not cause a local service to observe operations, but it should also not prevent upstream tracing-aware services from recording trace data based on their local sampler configuration. This requires samplers to support an additional undecided state in addition to observed, and not observed.

In practice this should prevent the `X-B3-Sampled` from being set on outgoing requests.

schlosna:
Is this resolved by https://github.com/palantir/tracing-java/pull/166?

carterkozak:
Not quite yet, I think we still want to update the sampler API to produce `UNDECIDED` unless specifically configured. This way requests from non-configurable clients sampled at a higher rate by reconfiguring the target server. Thoughts?

iamdanfox:
I started implementing this, but couldn't quite figure out how a TraceSampler would ever choose between UNDECIDED and DO_NOT_SAMPLE, so put together this approach: https://github.com/palantir/tracing-java/pull/170

carterkozak:
I think “undecided” should be used as the default sampler (would not impact WC services)

PR #158, opened 2019-05-21
[improvement] Set the default sampler to random with 1%

diogoholanda:
## Before this PR
The default sampler would sample all traces. I would claim this is not a great default and is somewhat error prone. It expects all consumers to override it to an appropriate value, with the penalty to failing to doing so being a high number of tracing logs that puts a lot of pressure in logging pipelines.

## After this PR
The default sampler is random with 1% probability of sampling.


carterkozak:
The default always-sampler has caused several issues for products I support getting spammed by custom java applications that do not override the default.

I support moving the default sampler from 100% to 1%. Implementing #32 would be a better solution, but requires an API change. @markelliot and @uschi2000 may have opinions and I'd like to hear their thoughts before proceeding.

ferozco:
👍 In support, Its unlikely anyone was intentionally using the current default

PR #163, opened 2019-05-29
[improvement] Allow setting observable flag when wrapping with new trace

yixunx:
## Before this PR

`Tracer.wrapWithNewTrace` and `Tracer.wrapWithAlternateTraceId` always initializes the new trace with `Observability.UNDECIDED`, and there is no easy way to force a particular trace to be sampled.

## After this PR

Added overloaded versions of `wrapWithNewTrace` and `wrapWithAlternateTraceId` that take an additional parameter to allow users to set whether the new trace should be observable.

This is useful to us because we have a service that makes requests to two separate backends, one of which is wrapped using `wrapWithNewTrace` so that it generates a separate trace, and we want the secondary trace to be sampled when the primary trace is sampled in order to compare the traces. This should become possible with this PR and https://github.com/palantir/tracing-java/pull/164.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -217,19 +217,24 @@ public final class Tracers {
         return wrapWithNewTrace(DEFAULT_ROOT_SPAN_OPERATION, delegate);
     }
 
+    public static <V> Callable<V> wrapWithNewTrace(String operation, Callable<V> delegate) {
+        return wrapWithNewTrace(operation, delegate, Observability.UNDECIDED);
+    }
+
     /**
      * Wraps the given {@link Callable} such that it creates a fresh {@link Trace tracing state} for its execution.
      * That is, the trace during its {@link Callable#call() execution} is entirely separate from the trace at
      * construction or any trace already set on the thread used to execute the callable. Each execution of the callable
      * will have a fresh trace. The given {@link String operation} is used to create the initial span.
      */
-    public static <V> Callable<V> wrapWithNewTrace(String operation, Callable<V> delegate) {
+    public static <V> Callable<V> wrapWithNewTrace(String operation, Callable<V> delegate,
+            Observability observability) {
         return () -> {
             // clear the existing trace and keep it around for restoration when we're done
             Optional<Trace> originalTrace = Tracer.getAndClearTraceIfPresent();
 
             try {
-                Tracer.initTrace(Observability.UNDECIDED, Tracers.randomId());
+                Tracer.initTrace(observability, Tracers.randomId());
                 Tracer.startSpan(operation);
                 return delegate.call();
             } finally {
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -281,21 +290,25 @@ public static Runnable wrapWithAlternateTraceId(String traceId, Runnable delegat
         return wrapWithAlternateTraceId(traceId, DEFAULT_ROOT_SPAN_OPERATION, delegate);
     }
 
+    public static Runnable wrapWithAlternateTraceId(String traceId, String operation, Runnable delegate) {
```
ferozco, line 293: mark this and the other compatibility methods as deprecated?

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -217,19 +217,24 @@ public static ScheduledExecutorService wrapWithNewTrace(String operation,
...
+
     /**
      * Wraps the given {@link Callable} such that it creates a fresh {@link Trace tracing state} for its execution.
      * That is, the trace during its {@link Callable#call() execution} is entirely separate from the trace at
      * construction or any trace already set on the thread used to execute the callable. Each execution of the callable
      * will have a fresh trace. The given {@link String operation} is used to create the initial span.
      */
-    public static <V> Callable<V> wrapWithNewTrace(String operation, Callable<V> delegate) {
+    public static <V> Callable<V> wrapWithNewTrace(String operation, Callable<V> delegate,
+            Optional<Boolean> isObservable) {
```
iamdanfox: this seems bonkers

ferozco: looks like the optional boolean is used as a tri-state where the options are: Sample, Don't Sample, Let the sampler decide whether to sample

yixunx: Yeah @iamdanfox if you are referring to the `Optional<Boolean>` then it's already used in the current tracing logic. I can add some comments to explain its purpose since I'm making it part of the public API.

tracing/src/main/java/com/palantir/tracing/Tracers.java
yixunx: done

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -210,26 +210,40 @@ public static ScheduledExecutorService wrapWithNewTrace(String operation,
...
-        return wrapWithNewTrace(DEFAULT_ROOT_SPAN_OPERATION, delegate);
+        return wrapWithNewTrace(DEFAULT_ROOT_SPAN_OPERATION, delegate, Optional.empty());
+    }
+
+    /**
+     * Deprecated.
+     *
+     * @deprecated Use {@link #wrapWithNewTrace(String, Callable, Optional)}
+     */
+    @Deprecated
```
iamdanfox: @ferozco I actually disagree that this should be deprecated.  It appears in _so many places_ in our code, and I think forcing everyone to make the decision about whether it should be sampled will actually result in more people messing things up.

ferozco: Happy to keep it around then :)

tracing/src/main/java/com/palantir/tracing/Tracers.java
iamdanfox: Given that we're adding a brand new method here, we have an opportunity to make a better API.   Even with documentation on the method, Optional<Boolean> is a pretty weird looking signature.

I'd propose something like:

```java
   enum Observability {
        SAMPLE,
        DO_NOT_SAMPLE,
        SAMPLER_DECIDES
    }
```

These probably aren't quite the perfect names yet, but I think they're preferrable to Optional<Boolean>!

yixunx: @iamdanfox That makes sense. I initially didn't do this because I didn't want to introduce a breaking change, but agree that it's much clearer with a enum instead of an optional. I opened a separate PR for the refactor: https://github.com/palantir/tracing-java/pull/166

yixunx:
Rebased onto develop to pick up #166. This is ready for review again.

ferozco:
👍

PR #164, opened 2019-05-29
[improvement] Expose Tracer.copyTrace

yixunx:
## Before this PR

There is no easy way to inspect the current trace without modifying the tracing state.

## After this PR

Exposed `Tracer.copyTrace` to allow access to the current tracing state.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -338,7 +338,7 @@ public final class Tracer {
     }
 
     /** Returns an independent copy of this thread's {@link Trace}. */
-    static Optional<Trace> copyTrace() {
+    public static Optional<Trace> copyTrace() {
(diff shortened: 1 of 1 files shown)


yixunx:
The use case is documented in https://github.com/palantir/tracing-java/pull/163.

iamdanfox (changes requested):
I think this is dangerous

yixunx:
@iamdanfox Can you give an example for why this is dangerous? I think it's fine if we only return a deep copy, but I might be missing some context here.

ferozco:
@iamdanfox what is your concern?

yixunx:
@iamdanfox 	Actually I just found out that there is already a `Tracer.isTraceObservable()` that does what we need, so we don't need this PR to support the use case mentioned #163. I do have another potential use case for exposing part of the current trace info, but I'll open a separate PR to discuss.

PR #166, opened 2019-06-03
[improvement] Refactor Tracer.initTrace to take enum instead of optional boolean

yixunx:
## Before this PR

`Tracer.initTrace` takes a slightly awkward approach to use an optional boolean to represent a tri-state (relevant discussion at https://github.com/palantir/tracing-java/pull/163#discussion_r289059477)

## After this PR

Added a new version of `Tracer.initTrace` that takes an enum instead of an optional boolean for the observability parameter. The old method that takes an optional is deprecated.

### DIFF
--- a/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
+++ b/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
@@ -102,7 +102,8 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
         }
     }
 
-    // Returns true iff the context contains a "1" X-B3-Sampled header, or absent if there is no such header.
+    // Force sample iff the context contains a "1" X-B3-Sampled header, force not sample if the header contains another
+    // non-empty value, or undecided if there is no such header or the header is empty.
     private static Observability getObservabilityFromHeader(ContainerRequestContext context) {
         String header = context.getHeaderString(TraceHttpHeaders.IS_SAMPLED);
         if (Strings.isNullOrEmpty(header)) {
--- a/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
+++ b/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
@@ -17,6 +17,7 @@
 package com.palantir.tracing.jersey;
 
 import com.google.common.base.Strings;
+import com.palantir.tracing.Observability;
 import com.palantir.tracing.Tracer;
 import com.palantir.tracing.Tracers;
 import com.palantir.tracing.api.Span;
@@ -66,10 +67,10 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
         // Set up thread-local span that inherits state from HTTP headers
         if (Strings.isNullOrEmpty(traceId)) {
             // HTTP request did not indicate a trace; initialize trace state and create a span.
-            Tracer.initTrace(hasSampledHeader(requestContext), Tracers.randomId());
+            Tracer.initTrace(getObservabilityFromHeader(requestContext), Tracers.randomId());
             Tracer.startSpan(operation, SpanType.SERVER_INCOMING);
         } else {
-            Tracer.initTrace(hasSampledHeader(requestContext), traceId);
+            Tracer.initTrace(getObservabilityFromHeader(requestContext), traceId);
             if (spanId == null) {
                 Tracer.startSpan(operation, SpanType.SERVER_INCOMING);
             } else {
@@ -102,12 +103,12 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
     }
 
     // Returns true iff the context contains a "1" X-B3-Sampled header, or absent if there is no such header.
-    private static Optional<Boolean> hasSampledHeader(ContainerRequestContext context) {
+    private static Observability getObservabilityFromHeader(ContainerRequestContext context) {
         String header = context.getHeaderString(TraceHttpHeaders.IS_SAMPLED);
         if (header == null) {
-            return Optional.empty();
+            return Observability.SAMPLER_DECIDES;
         } else {
-            return Optional.of(header.equals("1"));
+            return header.equals("1") ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE;
         }
     }
 }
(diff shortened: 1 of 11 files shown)


yixunx:
This is using the names @iamdanfox initially wrote in https://github.com/palantir/tracing-java/pull/163#discussion_r289944952. I think these look clear enough, but let me know if you come up with better names for this.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -62,32 +63,32 @@ private Tracer() {}
...
 
     /**
-     * Initializes the current thread's trace, erasing any previously accrued open spans. The new trace is {@link
-     * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent, if the {@link
-     * #setSampler configured sampler} returns true.
+     * Initializes the current thread's trace, erasing any previously accrued open spans.
      */
-    public static void initTrace(Optional<Boolean> isObservable, String traceId) {
-        setTrace(createTrace(isObservable, traceId));
+    public static void initTrace(Observability observability, String traceId) {
```
schlosna, line 104: Should we consider deprecating the version of this method that takes optional temporarily to maintain back-compat, or do we just want a hard break and roll forward?

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -102,12 +103,12 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
     // Returns true iff the context contains a "1" X-B3-Sampled header, or absent if there is no such header.
-    private static Optional<Boolean> hasSampledHeader(ContainerRequestContext context) {
+    private static Observability getObservabilityFromHeader(ContainerRequestContext context) {
         String header = context.getHeaderString(TraceHttpHeaders.IS_SAMPLED);
         if (header == null) {
-            return Optional.empty();
+            return Observability.SAMPLER_DECIDES;
         } else {
-            return Optional.of(header.equals("1"));
+            return header.equals("1") ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE;
```
schlosna: Happy this no longer allocates 😄

tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
```diff
@@ -81,14 +74,14 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
     // or absent if there is no such header.
-    private static Optional<Boolean> hasSampledHeader(HeaderMap headers) {
+    private static Observability getObservabilityFromHeader(HeaderMap headers) {
         String header = headers.getFirst(IS_SAMPLED);
         if (header == null) {
-            return Optional.empty();
+            return Observability.SAMPLER_DECIDES;
         } else {
             // No need to box the resulting boolean and allocate
             // a new Optional wrapper for each invocation.
```
schlosna: Can probably remove this comment

yixunx: done

tracing/src/main/java/com/palantir/tracing/Tracer.java
yixunx: This is the only place this is exposed, so should be pretty low-cost to maintain backcompat. I've added the old method back and marked it as deprecated.

tracing/src/main/java/com/palantir/tracing/Observability.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+public enum Observability {
+    SAMPLE,
+    DO_NOT_SAMPLE,
+    SAMPLER_DECIDES
```
uschi2000: maybe consider UNDECIDED

yixunx:
Think I've addressed all existing comments, and this is good for another review

tracing/src/main/java/com/palantir/tracing/Observability.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+public enum Observability {
+    SAMPLE,
+    DO_NOT_SAMPLE,
+    UNDECIDED
+}
```
carterkozak, line 37: Mind adding docs to this class?

Big fan of this :-)

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -102,12 +103,12 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
     }
 
     // Returns true iff the context contains a "1" X-B3-Sampled header, or absent if there is no such header.
-    private static Optional<Boolean> hasSampledHeader(ContainerRequestContext context) {
+    private static Observability getObservabilityFromHeader(ContainerRequestContext context) {
         String header = context.getHeaderString(TraceHttpHeaders.IS_SAMPLED);
         if (header == null) {
```
carterkozak: It's a slight behavior change, but we may want `Strings.isNullOrEmpty(header)` here

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -102,12 +103,12 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
-            return Optional.empty();
+            return Observability.UNDECIDED;
         } else {
-            return Optional.of(header.equals("1"));
+            return header.equals("1") ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE;
```
carterkozak: pedantic nit: would prefer `"1".equals(header)`, though it won't make a difference in practice here since we've already validated non-null

carterkozak (commented):
Generally lgtm, thanks for the deprecated backcompat methods.

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
dansanduleac: In favor of this, would lead to less potential confusion down the line.

carterkozak (approved):

PR #167, opened 2019-06-06
[improvement] use of factory method to avoid OpenSpan.Builder allocations

wenhoujx:
- keep builder() method for back-compat.

## Before this PR

flight recording of my java application shows `OpenSpan$Builder` has 3200 TLAB allocations amounts to 167.99 MB total. 
## After this PR

The static factory method `of` avoids `Builder` allocation.

### DIFF
--- a/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
@@ -33,6 +33,7 @@ public abstract class OpenSpan {
     /**
      * Returns a description of the operation for this event.
      */
+    @Value.Parameter
     public abstract String getOperation();
 
     /**
@@ -41,6 +42,7 @@ public abstract class OpenSpan {
      * Users of this class should not set this value manually in the builder, it is configured automatically when using
      * the {@link #builder()} static.
      */
+    @Value.Parameter
     public abstract long getStartTimeMicroSeconds();
 
     /**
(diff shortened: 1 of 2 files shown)


wenhoujx:
image

wenhoujx:
@carterkozak  @schlosna

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -139,19 +134,16 @@ public static OpenSpan startSpan(String operation) {
...
-                .spanId(Tracers.randomId())
-                .type(type);
-
         Trace trace = getOrCreateCurrentTrace();
         Optional<OpenSpan> prevState = trace.top();
+        final OpenSpan span;
         // Avoid lambda allocation in hot paths
         if (prevState.isPresent()) {
-            spanBuilder.parentSpanId(prevState.get().getSpanId());
+            span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.of(prevState.get().getSpanId()));
```
schlosna, line 142: I realize this is existing and we're stuck with this API, but the `Optional` allocations on hot paths make me sad.

schlosna (approved):
👍 LGTM, think someone else will need to +1

Created https://github.com/palantir/tracing-java/pull/168 to remove some other allocations from hot paths when we're not observing that popped in similar profiles, especially for wrapped executors & high frequency endpoints at low sample rates.

tracing/src/main/java/com/palantir/tracing/Tracer.java
wenhoujx: me sad too, do you think it's a good idea to create three methods to avoid the `Optional` wrapping? 
```
static OpenSpan.of( .......other params...... Optional<String> parentSpanId) ;
static OpenSpan.of( .......other params......String parentSpanId) ;
static OpenSpan.of( .......other params.....) ;

```

schlosna: I don't think its worth it as they'll still get created

wenhoujx: 👍

tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
```diff
@@ -77,6 +83,13 @@ public static Builder builder() {
                 .startClockNanoSeconds(System.nanoTime());
     }
 
+    /**
+     * Use this factory method to avoid allocate {@link Builder} in hot path.
+     */
+    public static OpenSpan of(String operation, String spanId, SpanType type, Optional<String> parentSpanId) {
```
iamdanfox, line 89: It's a bit of a shame that we have to make this public in order to access it within this library, as I really hope that no users will ever actually touch this class!

iamdanfox (approved):

PR #168, opened 2019-06-06
[improvement] Avoid trace ID and span allocations when not observing

schlosna:
## Before this PR

Related to #167 there is a lot of memory allocation, garbage collection overhead, and CPU overhead of random ID generation for trace IDs and spans when traces are not observable (i.e. either not explicitly defined as observable or not selected for sampling).

## After this PR

Avoid creating spans when the containing trace is not observable and lazily generate trace IDs only for observable traces to avoid the memory allocation, garbage collection overhead, and CPU overhead of random generation.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -30,8 +30,6 @@ import java.util.Optional;
  */
 public final class Trace {
 
-    private static final Trace NOOP = new Trace(false, "noop");
-
     private final Deque<OpenSpan> stack;
     private final boolean isObservable;
     private final String traceId;
@@ -44,15 +42,9 @@ public final class Trace {
         this.traceId = traceId;
     }
 
-    static Trace create(boolean isObservable, CharSequence traceId) {
-        if (isObservable) {
-            return new Trace(isObservable, traceId.toString());
-        }
-        return NOOP;
-    }
-
-    Trace(boolean isObservable, String traceId) {
-        this(new ArrayDeque<>(), isObservable, traceId);
+    static Trace create(boolean isObservable, String traceId) {
+        Deque<OpenSpan> deque = isObservable ? new ArrayDeque<>() : ImmutableEmptyDeque.instance();
+        return new Trace(deque, isObservable, traceId);
     }
 
     void push(OpenSpan span) {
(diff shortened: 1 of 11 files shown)


tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -30,24 +30,35 @@
...
     }
 
+    static Trace create(boolean isObservable, CharSequence traceId) {
+        if (isObservable) {
+            return new Trace(isObservable, traceId.toString());
+        }
+        return NOOP;
+    }
+
     Trace(boolean isObservable, String traceId) {
```
iamdanfox: Wanna mark this @VisibleForTesting and make `DeferredTracer` call your new static `create` method?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -30,24 +30,35 @@
...
         this.isObservable = isObservable;
         this.traceId = traceId;
     }
 
+    static Trace create(boolean isObservable, CharSequence traceId) {
+        if (isObservable) {
+            return new Trace(isObservable, traceId.toString());
+        }
+        return NOOP;
+    }
```
carterkozak: I think we still need to generate random trace IDs for non-observable traces, since the ID is applied to requests and logging MDC allowing us to load all the logs for an operation across services, even when the operation has not been sampled.

Fortunately traces are relatively uncommon, spans are created significantly more frequently, and the span ids are unimportant for unsampled traces.

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -30,24 +30,35 @@
  */
 public final class Trace {
 
+    private static final Trace NOOP = new Trace(false, "noop");
```
iamdanfox: Could we put a one-liner of javadoc on this guy to say that we just re-use this single instance for all non-observable traces?

tracing/src/main/java/com/palantir/tracing/Trace.java
iamdanfox: Mmm good point, setTrace will always call `MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());`, and we really do want those traceIds so we can correlate logs properly.

schlosna: yeah, probably need to remove the `lazyTraceId` then as there isn't a benefit if we're immediately going to generate the `String` form for MDC.

carterkozak: It may still be helpful for span IDs, which iirc use the same function. If we have 500 spans per trace, that may provide a sizable improvement.

carterkozak:
I think we still need to track span depth for non-sampled traces. When the root span is completed via `Tracer.completeSpan/fastCompleteSpan` the trace thread state is removed, along with the `traceId` MDC value. An approach I've considered is to use a different data type for sampled and unsampled traces, where unsampled traces have a traceId and a counter in place of the stack of spans. when the counter is decremented to zero, we can remove thread state.

schlosna:
closing in favor of #180

PR #170, opened 2019-06-07
[improvement] Don't impose DO_NOT_SAMPLE on downstream services

iamdanfox:
## Before this PR

If a service decides to start sampling a trace, it's important that all downstream service do the same to ensure we get a nice complete set of data in trace-log-viewer.

HOWEVER, if the TraceSampler decides _not_ to sample a trace, it currently imposes this DO_NOT_SAMPLE decision on all downstream services too by setting the X-B3-Sampled header to `0`

This means that theoretically if a single service early in a call graph has a low sampling rate, then it may unnecessarily reduce tracing information that all downstream services, even if they had plenty of capacity to record traces.

## After this PR

If a service decides not to observe a span, it no longer forces all the downstream services to also drop the span, and instead allows downstream samplers to decide.

Longer-term, I'd also quite like to make our trace sampling to be a target logging rate (e.g. up to 1000 lines/second) rather than just a constant percentage of incoming requests. This should mean that infrequently accessed services could just sample _every_ network request, thereby hopefully making life easier for developers.

## Possible downsides

- Rolling out this change will probably result in a higher proportion of traces getting sampled, as suddenly a request that passes through will have a chance to be sampled in every service, whereas previously the first failure would effectively blacklist that trace forever.
- I also don't really understand why you would want to mark a trace as explicitly DO_NOT_SAMPLE, so maybe I'm missing something here.

_Related to https://github.com/palantir/tracing-java/issues/32_

### DIFF
--- a/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
+++ b/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
@@ -46,8 +46,12 @@ public enum OkhttpTraceInterceptor implements Interceptor {
         OpenSpan span = Tracer.startSpan(spanName, SpanType.CLIENT_OUTGOING);
         Request.Builder tracedRequest = request.newBuilder()
                 .addHeader(TraceHttpHeaders.TRACE_ID, Tracer.getTraceId())
-                .addHeader(TraceHttpHeaders.SPAN_ID, span.getSpanId())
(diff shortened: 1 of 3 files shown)


ferozco:
As you noted this change will lead to a dramatic increase in observed traces. I worry that it may have negative impact on the signals processing pipeline since we are now sampling a request with probability `p = sum_i=0^n .98^i * .02` where n is the number of hops. For a request with 5 hops that translates to a sampling rate of 11.4%

carterkozak:
I think this makes it more difficult and dangerous to configure the trace rate because of the consequences for all downstream services.

iamdanfox:
Ok fair enough let's park this one

Issue #169, opened 2019-06-07
Add JMH benchmarks


schlosna:
## What happened?

It is difficult to quantify the impact of code changes during PR reviews of performance critical code paths.

## What did you want to happen?

Having a set of JMH benchmarks included with this project would make it easier to verify impact of optimizations such as #167 and #168 .

schlosna:
Closed by #179

PR #175, opened 2019-06-14
[improvement] add jmh to tracing-java

wenhoujx:
## Before this PR

hard to measure perf gain of optimization PR. 
## After this PR

we can use jmh to prove the perf gain. 

link to #169

### DIFF
--- a/tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracingBenchMark.java
+++ b/tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracingBenchMark.java
@@ -1,0 +1,50 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


wenhoujx:
@schlosna  @carterkozak  is there anything more to add?

tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracingBenchMark.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+
+@State(Scope.Thread)
+public class TracingBenchMark {
+    @Benchmark
+    public void benchmarkRandomId() {
+        Tracers.randomId();
```
carterkozak: We need to return the value from this function, otherwise JIT DCE will throw off the reported values.

gradle/jmh.gradle
```diff
@@ -0,0 +1,81 @@
+apply plugin: 'java'
+apply plugin: 'idea'
+
+project.sourceSets.create('jmh') {
+    java.srcDir 'src/jmh/java'
```
carterkozak: I think we can do this without the source-set hackery, I added similar benchmarks to auth-tokens a while back and I think it was simpler

wenhoujx: i removed this file, i keep jmh in the test path.

carterkozak:
Closing in favor of #176

PR #180, opened 2019-06-20
[improvement] Tracer spans are inexpensive for unsampled operations

carterkozak:
## Before this PR
Operations which are fast in the common case were infeasible to trace because regardless of the trace rate, span creation performance was worse than the operation being measured.

```
Benchmark                                                      (observability)  Mode  Cnt      Score       Error   Units
TracingBenchmark.nestedSpans                                            SAMPLE  avgt    3  39710.354 ± 36443.873   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                             SAMPLE  avgt    3   3425.430 ±  3215.533  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                        SAMPLE  avgt    3  41560.021 ±     0.434    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space                    SAMPLE  avgt    3   3444.571 ±  3116.221  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm               SAMPLE  avgt    3  41795.982 ±  7880.516    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space                SAMPLE  avgt    3      0.386 ±     2.219  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm           SAMPLE  avgt    3      4.741 ±    30.088    B/op
TracingBenchmark.nestedSpans:·gc.count                                  SAMPLE  avgt    3    137.000              counts
TracingBenchmark.nestedSpans:·gc.time                                   SAMPLE  avgt    3     87.000                  ms
TracingBenchmark.nestedSpans                                     DO_NOT_SAMPLE  avgt    3  19958.082 ±  6584.116   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                      DO_NOT_SAMPLE  avgt    3   3506.264 ±  1147.188  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                 DO_NOT_SAMPLE  avgt    3  21416.011 ±     0.230    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space             DO_NOT_SAMPLE  avgt    3   3547.447 ±  1525.303  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm        DO_NOT_SAMPLE  avgt    3  21666.731 ±  4251.611    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space         DO_NOT_SAMPLE  avgt    3      0.383 ±     0.331  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm    DO_NOT_SAMPLE  avgt    3      2.341 ±     1.300    B/op
TracingBenchmark.nestedSpans:·gc.count                           DO_NOT_SAMPLE  avgt    3    139.000              counts
TracingBenchmark.nestedSpans:·gc.time                            DO_NOT_SAMPLE  avgt    3     86.000                  ms
TracingBenchmark.nestedSpans                                         UNDECIDED  avgt    3  22570.008 ±  7874.344   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                          UNDECIDED  avgt    3   3590.635 ±  1268.870  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                     UNDECIDED  avgt    3  24792.658 ±    44.976    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space                 UNDECIDED  avgt    3   3674.586 ±  3668.504  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm            UNDECIDED  avgt    3  25360.986 ± 16430.980    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space             UNDECIDED  avgt    3      0.217 ±     2.148  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm        UNDECIDED  avgt    3      1.509 ±    15.297    B/op
TracingBenchmark.nestedSpans:·gc.count                               UNDECIDED  avgt    3     90.000              counts
TracingBenchmark.nestedSpans:·gc.time                                UNDECIDED  avgt    3     58.000                  ms
```

## After this PR
Added `Tracer.fastStartSpan` which can effectively no-op for unsampled operations.

```
Benchmark                                                      (observability)  Mode  Cnt      Score       Error   Units
TracingBenchmark.nestedSpans                                            SAMPLE  avgt    3  37969.436 ± 25282.797   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                             SAMPLE  avgt    3   3475.679 ±  2277.025  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                        SAMPLE  avgt    3  40360.062 ±     1.751    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space                    SAMPLE  avgt    3   3511.677 ±  2743.986  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm               SAMPLE  avgt    3  40771.767 ±  7164.960    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space                SAMPLE  avgt    3      0.384 ±     0.428  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm           SAMPLE  avgt    3      4.451 ±     2.688    B/op
TracingBenchmark.nestedSpans:·gc.count                                  SAMPLE  avgt    3    148.000              counts
TracingBenchmark.nestedSpans:·gc.time                                   SAMPLE  avgt    3     88.000                  ms
TracingBenchmark.nestedSpans                                     DO_NOT_SAMPLE  avgt    3    908.402 ±    79.755   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                      DO_NOT_SAMPLE  avgt    3   2820.275 ±   290.026  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                 DO_NOT_SAMPLE  avgt    3    784.001 ±     0.011    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space             DO_NOT_SAMPLE  avgt    3   2834.984 ±   293.831  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm        DO_NOT_SAMPLE  avgt    3    788.123 ±   162.926    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space         DO_NOT_SAMPLE  avgt    3      0.223 ±     1.492  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm    DO_NOT_SAMPLE  avgt    3      0.062 ±     0.416    B/op
TracingBenchmark.nestedSpans:·gc.count                           DO_NOT_SAMPLE  avgt    3    126.000              counts
TracingBenchmark.nestedSpans:·gc.time                            DO_NOT_SAMPLE  avgt    3     77.000                  ms
TracingBenchmark.nestedSpans                                         UNDECIDED  avgt    3   1254.746 ±   233.856   ns/op
TracingBenchmark.nestedSpans:·gc.alloc.rate                          UNDECIDED  avgt    3   2945.625 ±   514.073  MB/sec
TracingBenchmark.nestedSpans:·gc.alloc.rate.norm                     UNDECIDED  avgt    3   1130.854 ±    22.743    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space                 UNDECIDED  avgt    3   2993.199 ±  1125.922  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Eden_Space.norm            UNDECIDED  avgt    3   1149.047 ±   258.339    B/op
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space             UNDECIDED  avgt    3      0.279 ±     0.890  MB/sec
TracingBenchmark.nestedSpans:·gc.churn.PS_Survivor_Space.norm        UNDECIDED  avgt    3      0.107 ±     0.355    B/op
TracingBenchmark.nestedSpans:·gc.count                               UNDECIDED  avgt    3    135.000              counts
TracingBenchmark.nestedSpans:·gc.time                                UNDECIDED  avgt    3     81.000                  ms
```

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -20,6 +20,7 @@ import static com.palantir.logsafe.Preconditions.checkArgument;
 import static com.palantir.logsafe.Preconditions.checkState;
 
 import com.google.common.base.Strings;
+import com.google.errorprone.annotations.CheckReturnValue;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tracing.api.OpenSpan;
@@ -45,7 +46,9 @@ public abstract class Trace {
     /**
      * Opens a new span for this thread's call trace, labeled with the provided operation and parent span. Only allowed
      * when the current trace is empty.
+     * If the return value is not used, prefer {@link #fastStartSpan(String, String, SpanType)}}.
      */
+    @CheckReturnValue
     final OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
         checkState(isEmpty(), "Cannot start a span with explicit parent if the current thread's trace is non-empty");
         checkArgument(!Strings.isNullOrEmpty(parentSpanId), "parentSpanId must be non-empty");
@@ -58,6 +61,7 @@ public abstract class Trace {
      * Opens a new span for this thread's call trace, labeled with the provided operation.
      * If the return value is not used, prefer {@link #fastStartSpan(String, SpanType)}}.
      */
+    @CheckReturnValue
     final OpenSpan startSpan(String operation, SpanType type) {
         Optional<OpenSpan> prevState = top();
         final OpenSpan span;
@@ -124,11 +128,13 @@ public abstract class Trace {
         }
 
         @Override
+        @SuppressWarnings("ResultOfMethodCallIgnored") // Sampled traces cannot optimize this path
         void fastStartSpan(String operation, String parentSpanId, SpanType type) {
             startSpan(operation, parentSpanId, type);
         }
 
         @Override
+        @SuppressWarnings("ResultOfMethodCallIgnored") // Sampled traces cannot optimize this path
         void fastStartSpan(String operation, SpanType type) {
             startSpan(operation, type);
         }
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -17,6 +17,7 @@
 package com.palantir.tracing;
 
 import static com.palantir.logsafe.Preconditions.checkArgument;
+import static com.palantir.logsafe.Preconditions.checkState;
 
 import com.google.common.base.Strings;
 import com.palantir.logsafe.SafeArg;
@@ -41,7 +42,45 @@ public abstract class Trace {
         this.traceId = traceId;
     }
 
-    abstract void startSpan(String operation, SpanType type);
+    /**
+     * Opens a new span for this thread's call trace, labeled with the provided operation and parent span. Only allowed
+     * when the current trace is empty.
+     */
+    final OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
+        checkState(isEmpty(), "Cannot start a span with explicit parent if the current thread's trace is non-empty");
+        checkArgument(!Strings.isNullOrEmpty(parentSpanId), "parentSpanId must be non-empty");
+        OpenSpan span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.of(parentSpanId));
+        push(span);
+        return span;
+    }
+
+    /**
+     * Opens a new span for this thread's call trace, labeled with the provided operation.
+     * If the return value is not used, prefer {@link #fastStartSpan(String, SpanType)}}.
+     */
+    final OpenSpan startSpan(String operation, SpanType type) {
+        Optional<OpenSpan> prevState = top();
+        final OpenSpan span;
+        // Avoid lambda allocation in hot paths
+        if (prevState.isPresent()) {
+            span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.of(prevState.get().getSpanId()));
+        } else {
+            span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.empty());
+        }
+
+        push(span);
+        return span;
+    }
+
+    /**
+     * Like {@link #startSpan(String, String, SpanType)}, but does not return an {@link OpenSpan}.
+     */
+    abstract void fastStartSpan(String operation, String parentSpanId, SpanType type);
+
+    /**
+     * Like {@link #startSpan(String, SpanType)}, but does not return an {@link OpenSpan}.
+     */
+    abstract void fastStartSpan(String operation, SpanType type);
 
     abstract void push(OpenSpan span);
 
@@ -85,16 +124,13 @@ public abstract class Trace {
         }
 
         @Override
-        void startSpan(String operation, SpanType type) {
-            Optional<OpenSpan> prevState = top();
-            final OpenSpan span;
-            // Avoid lambda allocation in hot paths
-            if (prevState.isPresent()) {
-                span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.of(prevState.get().getSpanId()));
-            } else {
-                span = OpenSpan.of(operation, Tracers.randomId(), type, Optional.empty());
-            }
-            push(span);
+        void fastStartSpan(String operation, String parentSpanId, SpanType type) {
+            startSpan(operation, parentSpanId, type);
+        }
+
+        @Override
+        void fastStartSpan(String operation, SpanType type) {
+            startSpan(operation, type);
         }
 
         @Override
@@ -147,7 +183,12 @@ public abstract class Trace {
         }
 
         @Override
-        void startSpan(String operation, SpanType type) {
+        void fastStartSpan(String operation, String parentSpanId, SpanType type) {
+            fastStartSpan(operation, type);
+        }
+
+        @Override
+        void fastStartSpan(String operation, SpanType type) {
             depth++;
         }
 
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -192,7 +192,7 @@ public abstract class Trace {
         private Unsampled(int numberOfSpans, String traceId) {
             super(traceId);
             this.numberOfSpans = numberOfSpans;
-            validateDepth();
+            validateNumberOfSpans();
         }
 
         private Unsampled(String traceId) {
@@ -221,7 +221,7 @@ public abstract class Trace {
 
         @Override
         Optional<OpenSpan> pop() {
-            validateDepth();
+            validateNumberOfSpans();
             if (numberOfSpans > 0) {
                 numberOfSpans--;
             }
@@ -230,7 +230,7 @@ public abstract class Trace {
 
         @Override
         boolean isEmpty() {
-            validateDepth();
+            validateNumberOfSpans();
             return numberOfSpans <= 0;
         }
 
@@ -245,7 +245,7 @@ public abstract class Trace {
         }
 
         /** Internal validation, this should never fail because {@link #pop()} only decrements positive values. */
-        private void validateDepth() {
+        private void validateNumberOfSpans() {
             if (numberOfSpans < 0) {
                 throw new SafeIllegalStateException("Unexpected negative numberOfSpans",
                         SafeArg.of("numberOfSpans", numberOfSpans));
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -240,7 +240,8 @@ public abstract class Trace {
         /** Internal validation, this should never fail because {@link #pop()} only decrements positive values. */
         private void validateDepth() {
             if (numberOfSpans < 0) {
-                throw new SafeIllegalStateException("Unexpected negative numberOfSpans", SafeArg.of("numberOfSpans", numberOfSpans));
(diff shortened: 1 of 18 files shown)


carterkozak:
Still WIP, some refactors I need to make to reduce code duplication.

schlosna (approved):
👍 the 4x speedup and 4x reduction in allocations for common `UNDECIDED` case and 10x improvements for `DO_NOT_SAMPLE` are excellent 😄

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +33,209 @@
...
+        return span;
     }
 
-    Optional<OpenSpan> top() {
-        return stack.isEmpty() ? Optional.empty() : Optional.of(stack.peekFirst());
-    }
+    /**
+     * Like {@link #startSpan(String, String, SpanType)}, but does not return an {@link OpenSpan}.
+     */
+    abstract void fastStartSpan(String operation, String parentSpanId, SpanType type);
```
markelliot, line 89: perhaps what's warranted here is just a brand new tracing implementation? I don't think continuing to hack the current implementation even further.

I'm also perplexed by why this would be the primary target for perf improvement, we do lots of expensive processing throughout the stack of calls across our services, and while Tracing is a constant overhead it's likely dwarfed by serde costs.

carterkozak: > perhaps what's warranted here is just a brand new tracing implementation?

Sure, there are several problems we could solve in a new tracing implementation, I would be happy to help with that project when I have cycles, but it's not something that I can drive at the moment.

> I'm also perplexed by why this would be the primary target for perf improvement, we do lots of expensive processing throughout the stack of calls across our services, and while Tracing is a constant overhead it's likely dwarfed by serde costs.

That is certainly true of the standard conjure service stack, however this change is not targeting problems with conjure services. We have an internal application with unusually high throughput and low latency requirements, built on our service library (for configuration, status, logging, metrics, etc) where interoperability with the standard stack allows us to avoid reinventing these components.

carterkozak:
> 4x speedup and 4x reduction in allocations for common UNDECIDED case

This sounded lower than expected, the benchmark was set to 10% sample rate instead of 1% for the `UNDECIDED` case. I have updated the main comment with the more accurate numbers.

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +33,209 @@
...
         this.traceId = traceId;
     }
 
-    Trace(boolean isObservable, String traceId) {
-        this(new ArrayDeque<>(), isObservable, traceId);
+    /**
+     * Opens a new span for this thread's call trace, labeled with the provided operation and parent span. Only allowed
+     * when the current trace is empty.
+     */
+    final OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
```
iamdanfox, line 59: Is worth adding an error-prone `@CheckReturnValue` annotation to these methods, and recommending `fastStartSpan` if you don't actually need the OpenSpan object?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +33,209 @@
  * Represents a trace as an ordered list of non-completed spans. Supports adding and removing of spans. This class is
  * not thread-safe and is intended to be used in a thread-local context.
  */
-public final class Trace {
+public abstract class Trace {
```
iamdanfox, line 44: Not 100% sure, but I think this might be an ABI break.

A quick search internally suggested that nobody is actually importing the `com.palantir.tracing.Trace` class - perhaps we could even make it package-private

carterkozak: It’s exposed by Tracer.getAndClearTrace, however all methods are package private — it is only available as a marker

tracing/src/main/java/com/palantir/tracing/Trace.java
carterkozak: I did that on the public APIs, I can add annotations here, though the sampled tracer delegates internally and would need to suppress. Probably for the best in order to avoid internal misuse, I’ll update

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +34,214 @@
...
+    abstract void fastStartSpan(String operation, SpanType type);
 
-    boolean isEmpty() {
-        return stack.isEmpty();
-    }
+    abstract void push(OpenSpan span);
+
+    abstract Optional<OpenSpan> top();
+
+    abstract Optional<OpenSpan> pop();
```
iamdanfox, line 100: I find the contract of these two methods a bit confusing... normally an Optional.empty means an element was not present, but in this case the UnsampledTrace _always_ returns Optional.empty?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +34,214 @@
...
+        public String toString() {
+            return "Trace{stack=" + stack + ", isObservable=true, traceId='" + getTraceId() + "'}";
+        }
     }
 
-    @Override
-    public String toString() {
-        return "Trace{stack=" + stack + ", isObservable=" + isObservable + ", traceId='" + traceId + "'}";
+    private static final class Unsampled extends Trace {
+        private int depth;
```
iamdanfox: could we rename this `s/depth/numberOfSpans/`, maybe also a little comment to explain that the whole point of an unsampled trace is that it'll never notify SpanObservers, we don't need to store any of the span names that users provide

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +34,214 @@
...
+        }
+
+        @Override
+        Optional<OpenSpan> top() {
+            return Optional.empty();
+        }
+
+        @Override
+        Optional<OpenSpan> pop() {
+            validateDepth();
```
iamdanfox: Hmmm in the Sampled class, calling pop() repeatedly on an empty trace just returns Optional.empty over and over again

In this case though, it will throw.  Should we make these behave the same? or is this difference important?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -28,62 +34,214 @@
...
+            if (depth > 0) {
+                depth--;
+            }
+            return Optional.empty();
+        }
+
+        @Override
+        boolean isEmpty() {
+            validateDepth();
```
iamdanfox: In theory this call should be safe to delete right? As `pop()` is the only method that can decrement `depth`?

tracing/src/main/java/com/palantir/tracing/Trace.java
carterkozak: That is correct, this PR introduces a subtle change in behavior where completing an unsampled span (even using the slow complete span method) no longer returns a present value, because we know it is not recorded (in fact, sampling an unsampled span would be a bug).

Our tracing API is missing clean isolation between producers and consumers -- if I am creating spans, I don't need to know whether or not they are sampled, or what the span IDs/timing looks like, just how to start and stop them. Conceptually there's no reason to consume these values because that functionality is a concern of the `SpanObserver`.

The contract is defined by https://github.com/palantir/tracing-java/blob/42aafc8203a96be53ee52794f058d19d54db24fa/tracing/src/main/java/com/palantir/tracing/Tracer.java#L182-L188

There's a lot going on in that comment: "returns the current span (if it exists)". If we're not sampled, we don't really have a current span because we didn't record one. We have to track how many we would have created in the sampled case, but it does not violate the contract to return an empty value. This is reinforced by the next statement: "notifies all subscribers about the completed span", because subscribers are only notified if spans are sampled.

tracing/src/main/java/com/palantir/tracing/Trace.java
carterkozak: pop validates that the value is positive, this should never throw unless we write a bug. I can add a comment explaining that, or remove the validation.

tracing/src/main/java/com/palantir/tracing/Trace.java
carterkozak: Yep, validateDepth only exists to make sure I didn't write a bug :-)

ferozco:
👍

tracing/src/main/java/com/palantir/tracing/Trace.java
raiju: Hey, this is a very surprising PR. It is very much a break in behavior, and excusing it on a possible implied reading of the javadoc is not reasonable. This was done as a minor version bump (although given our current tooling a major version bump wouldn't have made a huge difference).

> Conceptually there's no reason to consume these values because that functionality is a concern of the SpanObserver.

If that were true we would only ever be using `fastCompleteSpan()`. As that isn't true, this is not an implementation detail but an API contract.

PR #202, opened 2019-07-24
Add initTrace javadoc regarding trace lifetime

JacekLach:
## Before this PR
It is very unclear that a traceid registered with `.initTrace` will be cleared as soon as the first span is closed.

## After this PR
Javadoc includes information about trace lifetime

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -100,6 +100,9 @@ public final class Tracer {
 
     /**
      * Initializes the current thread's trace, erasing any previously accrued open spans.
+     *
+     * The new trace will be persistent for the duration of the next span opened on it via
+     * {@link #startSpan}, and will be discarded when that span is completed or discarded.
      */
     public static void initTrace(Observability observability, String traceId) {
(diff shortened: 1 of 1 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -100,6 +100,9 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
 
     /**
      * Initializes the current thread's trace, erasing any previously accrued open spans.
+     *
+     * The new trace will be persistent for the duration of the next span opened on it via
+     * {@link #startSpan}, and will be discarded when that span is completed or discarded.
```
carterkozak, line 105: Instead of documenting effects, perhaps we should document how this is meant to be used correctly? This is only supposed to be called just before starting a new span, and only to set trace state when a request is received.

JacekLach: Yeah, there was some slack discussion on this being a fairly internal API. Happy to see other fixes, including possibly just changing this method to take a span name and initialise the span immediately, this was just a quick fix for something that bit me.

PR #209, opened 2019-07-31
Remove jackson from api

mtuer:
## After this PR
==COMMIT_MSG==
Reduces the dependency footprint of the tracing api by removing
Jackson. This dep is used on one class to add JsonDeserialization,
but is unused.
==COMMIT_MSG==

## Possible downsides?

Anyone that attempts to deserialize Spans would no longer be able to do so.

### DIFF
--- a/tracing-api/build.gradle
+++ b/tracing-api/build.gradle
@@ -1,9 +1,6 @@
 apply from: "$rootDir/gradle/publish-jar.gradle"
 
 dependencies {
-    compile "com.fasterxml.jackson.core:jackson-databind"
-    implementation 'com.fasterxml.jackson.core:jackson-annotations'
-
     testCompile "junit:junit"
     testCompile "org.assertj:assertj-core"
     testCompile "org.mockito:mockito-core"
--- a/tracing-api/src/main/java/com/palantir/tracing/api/Span.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/Span.java
@@ -16,8 +16,6 @@
 
 package com.palantir.tracing.api;
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.Map;
 import java.util.Optional;
 import org.immutables.value.Value;
@@ -25,8 +23,6 @@ import org.immutables.value.Value;
 /**
  * A value class representing a completed Span, see {@link OpenSpan} for a description of the fields.
  */
-@JsonDeserialize(as = ImmutableSpan.class)
-@JsonSerialize(as = ImmutableSpan.class)
 @Value.Immutable
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
 public abstract class Span {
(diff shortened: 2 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tracing-java](https://github.com/palantir/tracing-java), @mtuer! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

tracing-api/src/main/java/com/palantir/tracing/api/Span.java
```diff
@@ -16,17 +16,13 @@
...
 import org.immutables.value.Value;
 
 /**
  * A value class representing a completed Span, see {@link OpenSpan} for a description of the fields.
  */
-@JsonDeserialize(as = ImmutableSpan.class)
-@JsonSerialize(as = ImmutableSpan.class)
 @Value.Immutable
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
 public abstract class Span {
```
carterkozak, line 28: I don't know why this included jackson annotations to begin with other than copy/paste from another immutables type. @uschi2000, @markelliot, @ellisjoe are you aware of anything that intentionally deserializes tracing-api spans from json?

uschi2000: Forgot the exact details, but I suspect that the initial tracing implementation may have serialized this class directly instead of going via `ZipkinCompatSpan` in `AsyncSlf4jSpanObserver`. I am not aware of anyone serializing or deserializing via this class, but I also don't have exhaustive knowledge of the world, obviously. It's *probably* fine to remove, but I can't make a certain determination.

tracing-api/build.gradle
```diff
@@ -1,9 +1,6 @@
 apply from: "$rootDir/gradle/publish-jar.gradle"
 
 dependencies {
-    compile "com.fasterxml.jackson.core:jackson-databind"
-    implementation 'com.fasterxml.jackson.core:jackson-annotations'
```
iamdanfox, line 5: What would happen if we made these `compileOnly`, so that they don't appear in the `compile` or `runtime` sections of the POM? I think consumers who had jackson on the classpath would still be able to work?

carterkozak: It would avoid breaking consumers which serialize Span to json, but I'm curious if that's something anyone does, and if so, is it something that they should be able to do?

I think the more likely break is excavator failures when tracing-api no longer provides jackson-databind, but it's only likely to affect a few libraries with relatively few dependencies. The exact dependencies plugin protects consumers from this, and it's easy to fix, so it's not something I would optimize on.

carterkozak:
I'm happy with this, if it causes problems we can revert. I'll plan to merge and release tomorrow if there are no objections.

iamdanfox (approved):

carterkozak:
👍

carterkozak:
Looks like our required checks are out of date

iamdanfox:
Maybe just merge in develop?

carterkozak:
@mtuer if you rebase this should take care of itself. If not I can take a look.

PR #235, opened 2019-08-23
Test-utils to visualize spans

iamdanfox:
## Before this PR

I sometimes find it hard to intuitively imagine what spans look like, especially when we're talking about async stuff. We also don't have a good way of _testing_ spans in our libraries (e.g. conjure-java-runtime).

## After this PR
==COMMIT_MSG==
Tests can now be annotated `@TestTracing` to set up snapshot testing.
==COMMIT_MSG==

## Possible downsides / known limitations

- we don't currently serialize the 'type' (i.e. LOCAL / SERVER_OUTGOING), so can't make assertions based on it
- if users call `Tracer.setSampler` part way through the tests (which WC does), they'll get incomplete info in the tests
- running JUnit5 tests concurrently will result in very strange looking HTML reports
- when snapshotting is enabled, renaming test classes/methods could result in extraneous trace files left around
- we're not actually using the [SLS trace.1 format](https://github.com/palantir/witchcraft-api/blob/master/witchcraft-logging-api/src/main/conjure/witchcraft-logging-api.yml#L217)


tracing/src/test/java/com/palantir/tracing/HtmlSpanObserver.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public final class HtmlSpanObserver implements SpanObserver {
```
ferozco: Could we make this an extension/rule so that it automatically registers and outputs the contents?

iamdanfox: On it

iamdanfox:
I've tried this on c-j-r which is a bit weird because many tests don't actually neatly close a trace. I think it gives reasonable output.

carterkozak:
Have you considered shipping this functionality as an option for `slslog`?

tracing-test-utils/src/main/java/com/palantir/tracing/SpanRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+                .put(TimeUnit.MICROSECONDS, TimeUnit.MILLISECONDS)
+                .put(TimeUnit.MILLISECONDS, TimeUnit.SECONDS)
+                .build();
+
+        ImmutableMap<TimeUnit, String> abbreviation = ImmutableMap.<TimeUnit, String>builder()
+                .put(TimeUnit.NANOSECONDS, "ns")
+                .put(TimeUnit.MICROSECONDS, "micros")
+                .put(TimeUnit.MILLISECONDS, "ms")
+                .put(TimeUnit.SECONDS, "s")
+                .build();
```
carterkozak: nit: These maps should be static constants, no reason to create new maps each time we execute `renderDuration`

tracing-test-utils/src/main/java/com/palantir/tracing/SpanRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+
+@SuppressWarnings({"PreferSafeLoggableExceptions", "Slf4jLogsafeArgs"}) // test-lib, no need for SafeArgs
+final class SpanRenderer implements SpanObserver {
+
+    private static final Logger log = LoggerFactory.getLogger(SpanRenderer.class);
+    private final Collection<Span> allSpans = new ArrayBlockingQueue<>(1000);
+
+    @Override
+    public void consume(Span span) {
+        allSpans.add(span);
```
carterkozak: The SpanObserver rendering engine coupling feels odd. I think it would make sense if we could filter out events or process data as we consume it, but as is, we might as well take an `Collection<Span>` ctor argument, and build a SpanObserver separately.

tracing-test-utils/src/main/java/com/palantir/tracing/RenderTracingRule.java
```diff
@@ -0,0 +1,47 @@
...
+    public Statement apply(
+            Statement base, Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                try {
+                    Tracer.setSampler(AlwaysSampler.INSTANCE);
+                    Tracer.subscribe("RenderTracingRule", renderer);
+
+                    base.evaluate();
```
carterkozak, line 41: We may want to clear any existing span if present, and use a CloseableTracer around `base.evaluate()` with the test name.
Possibly worth a flag in the constructor because it does change tracing state, which may not always be what we want , especially when we're testing tracing.

tracing-test-utils/src/main/java/com/palantir/tracing/RenderTracingRule.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.palantir.tracing;
+
+import java.nio.file.Path;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public final class RenderTracingRule implements TestRule {
+    private final SpanRenderer renderer = new SpanRenderer();
```
carterkozak: SpanRenderer is stateful, I think we want to create this inside `evaluate()`.

tracing-test-utils/src/main/java/com/palantir/tracing/RenderTracingRule.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public Statement apply(
+            Statement base, Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                try {
+                    Tracer.setSampler(AlwaysSampler.INSTANCE);
+                    Tracer.subscribe("RenderTracingRule", renderer);
```
carterkozak: Unexpected things can happen if this rule is applied twice.
```suggestion
                    String subscriptionName = "RenderTracingRule" + UUID.randomUUID();
                    Tracer.subscribe("RenderTracingRule", renderer);
```

carterkozak (commented):
A lot of folks would love to have this in production environments that don't have the resources for the full caelum pipeline. We may want to move the rendering code from `tracing-test-utils` into a separate `tracing-visualization` package where it can be reused from a CLI.

tracing-test-utils/src/main/java/com/palantir/tracing/SpanRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+    void output(String displayName, Path path) {
+        TimeBounds bounds = bounds(allSpans);
+
+        Map<String, List<Span>> spansByTraceId = allSpans.stream()
+                .collect(Collectors.groupingBy(Span::getTraceId));
+
+        Map<String, AnalyzedSpans> analyzedByTraceId = Maps.transformValues(spansByTraceId, SpanRenderer::analyze);
+
+        HtmlFormatter formatter = new HtmlFormatter(bounds);
+        StringBuilder sb = new StringBuilder();
```
carterkozak: Can we operate on a Writer/PrintWriter instead of buffering the full output on heap twice?

tracing-test-utils/src/main/java/com/palantir/tracing/SpanRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+                    AnalyzedSpans analysis = entry.getValue();
+                    formatter.renderAllSpansForOneTraceId(entry.getKey(), analysis, sb);
+                });
+
+        formatter.rawSpanJson(allSpans, sb);
+
+        try {
+            Files.write(
+                    path,
+                    sb.toString().getBytes(StandardCharsets.UTF_8));
```
carterkozak: `sb` contains the full formatted output on heap
`sb.toString()` creates a second copy
`getBytes` create a third copy

tracing-test-utils/src/main/java/com/palantir/tracing/SpanRenderer.java
```diff
@@ -0,0 +1,382 @@
...
+                // Path file = new HtmlFormatter(bounds).emitToTempFile(analysis.orderedSpans());
+                // System.out.println("HTML span visualization: " + file);
+
+                // emit ASCII
+                AsciiFormatter ascii = new AsciiFormatter(bounds);
+                for (Span span : analysis.orderedSpans()) {
+                    System.out.println(ascii.formatSpan(span));
+                }
+            }
+        });
```
carterkozak: Should we have a way to opt in/out of ASCII or HTML output?

tracing-test-utils/src/main/java/com/palantir/tracing/RenderTracingRule.java
iamdanfox: When rendering/analyzing, we now always guarantee there is exactly one root span (by creating a fake one if necessary)

tracing-test-utils/src/main/java/com/palantir/tracing/RenderTracingRule.java
carterkozak: If we create a new span for each test, we don't have to worry about creating fake root spans, and it allows us to check the traceId to filter only spans in the current test. Otherwise parallel tests will report spans from tests, since Tracer observers use a static binding.

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.support.AnnotationSupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class TestTracingExtension implements BeforeEachCallback, AfterEachCallback {
```
CRogers, line 39: why `BeforeEachCallback` rather than `BeforeTestExecution`?

ferozco: It seems like BeforeTestExecution was causing the spans within the setup method to also be captured

PR #247, opened 2019-08-30
tweaks to ckozak's DetachedSpan & demo usage

iamdanfox:
## Before this PR

In conjure-java-runtime and dialogue, we want to be able to measure a span of time beginning on one thread and ending on another (e.g. backoffs or client-side concurrency limits).  We do have an `AsyncTracer` class already, but this makes two spans: `<operation>-enqueue` and `<operation>-run`.

## After this PR
~A new `CrossThreadSpan` class allows measuring operations that begin on one thread and end on another.~

~I'm not expecting this to be used by regular service authors much, but hoping we can benefit from it in our libraries.~

image

## Possible downsides?

_TODO_


tracing/src/main/java/com/palantir/tracing/CrossThreadSpan.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @CheckReturnValue
+    public static CrossThreadSpan startSpan(String operation) {
+        Trace trace = Tracer.hasTraceId()
+                ? Tracer.getOrCreateCurrentTrace()
+                : Tracer.createTrace(Observability.UNDECIDED, Tracers.randomId());
+
+        Optional<String> parentSpanId = trace.top().map(OpenSpan::getSpanId);
+        OpenSpan openSpan1 = OpenSpan.of(
```
markelliot: `openSpan1` -> `openSpan`?

markelliot: oh i see, that'd conflict with the global, nvm

tracing/src/main/java/com/palantir/tracing/CrossThreadSpan.java
```diff
@@ -0,0 +1,146 @@
...
+     * Finish the cross-thread span and start a new (thread local span) with the given operation name.
+     * NOTE: this is destructive, because any existing threadlocal tracing state is thrown away.
+     */
+    @MustBeClosed
+    public CloseableTracer sibling(String nextOperationName) {
+        OpenSpan inProgress = consumeOpenSpan("Can't call CrossThreadSpan#sibling if span has already been closed");
+
+        Map<String, String> metadata = Collections.emptyMap();
+        Span span = Tracer.toSpan(inProgress, metadata, traceId);
+        Tracer.notifyObservers(span);
```
markelliot: it'd be nice to pull lines 85-89 and 70-76 into a common private function

tracing/src/main/java/com/palantir/tracing/CrossThreadSpan.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // we throw away the existing thread local state and replace it with our view of the world
+        Tracer.clearCurrentTrace();
+        Tracer.setTrace(Trace.of(observable, traceId));
```
markelliot: I think maybe I'd wrap up the clearCurrentTrace+setTrace into a couple a private method `replaceTraceState` that takes no args, you can reuse it in `child`, too

tracing/src/main/java/com/palantir/tracing/CrossThreadSpan.java
```diff
@@ -0,0 +1,146 @@
...
+
+        OpenSpan inProgress = accessOpenSpan("Can't create child after terminate has been called");
+
+        Tracer.setTrace(Trace.of(observable, traceId));
+
+        return CloseableTracer.startSpan(nextOperationName, Optional.of(inProgress.getSpanId()), SpanType.LOCAL);
+    }
+
+    /**
+     * Unusual method - returns a new {@link CrossThreadSpan} parented to this instance. Only necessary if you want
```
markelliot: would be helpful to include an example in the docs of when to use this, it's not obvious how to select `child` or this method, both result in a parented CrossThreadSpan?

markelliot:
@jellis curious if this works for your needs?

iamdanfox:
Ok so this is just gonna land into the existing DetachedSpan branch - I think it will let us do all the tracing we want in c-j-r. Hope the names look OK from your perspective @carterkozak?

Definitely still want to explore Mark's new internal representation ideas, but don't want to block this PR on that (as it should be easy to polyfill later if we need to).

tracing/src/main/java/com/palantir/tracing/CloseableSpan.java
```diff
@@ -21,10 +21,10 @@
 /**
  * Closeable marker around a tracing span operation. This object should be used in a try/with block.
  */
-public interface SpanToken extends Closeable {
+public interface CloseableSpan extends Closeable {
```
carterkozak, line 24: 👍 I like this better.

carterkozak:
👍

ferozco:
👍

PR #249, opened 2019-09-04
Tracer#getTraceMetadata [into ckozak/detached_tracing_api]

iamdanfox:


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -89,6 +90,32 @@ private static boolean shouldObserve(Observability observability) {
...
+                    .spanId(openSpan.getSpanId())
+                    .parentSpanId(openSpan.getParentSpanId())
+                    .originatingSpanId(trace.getOriginatingSpanId())
+                    .traceId(trace.getTraceId())
+                    .build();
+        } else {
+            // In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId
+            // stack, so we just make one up (just in time). This matches the behaviour of Tracer#startSpan.
+
+            // 🌶🌶🌶 this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
```
```suggestion
            // n.b. this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
```

carterkozak:
👍

PR #262, opened 2019-09-11
Fix #177: Add APIs for Supplier<String> operation names

### LINKED ISSUE #177: Tracer.startSpan should support overloads to parameterize messages
## What happened?

We do a great deal of unnecessary string concatenation in span names which can be completely avoided for unsampled traces:

https://github.com/palantir/tracing-java/blob/c63567eacecc677c63db0f146fcd2f34493f2ffa/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java#L62

https://github.com/palantir/tracing-java/blob/c63567eacecc677c63db0f146fcd2f34493f2ffa/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java#L42

Slf4j style overloads to construct trace strings only for sampled operations

carterkozak:
## Before this PR
Hacks scattered across consumers to avoid expensive string work when the name will not be used, e.g. https://github.com/palantir/tritium/blob/880c6da14f7d89644fce935a5fe0fb8e080b4586/tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java#L97-L103

## After this PR
`CloseableTracer.startSpan(() -> "Foo: " + bar)`
`Tracer.fastStartSpan(() -> "Foo: " + bar)`
==COMMIT_MSG==
Add Supplier<String> operation name APIs for lazy operation name evaluation
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
@@ -17,6 +17,7 @@
 package com.palantir.tracing;
 
 import com.palantir.tracing.api.SpanType;
+import java.util.function.Supplier;
 
 /**
  * Wraps the {@link Tracer} methods in a closeable resource to enable the usage of the try-with-resources pattern.
@@ -40,12 +41,30 @@ public final class CloseableTracer implements AutoCloseable {
     }
 
     /**
+     * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
+     */
+    public static CloseableTracer startSpan(Supplier<String> operation) {
+        return startSpan(operation, SpanType.LOCAL);
+    }
+
+    /**
      * Opens a new span for this thread's call trace with the provided {@link SpanType},
      * labeled with the provided operation.
      *
      * If you need to a span that may complete on another thread, use {@link DetachedSpan#start} instead.
      */
     public static CloseableTracer startSpan(String operation, SpanType spanType) {
+        Tracer.fastStartSpan(operation, spanType);
+        return INSTANCE;
+    }
+
+    /**
+     * Opens a new span for this thread's call trace with the provided {@link SpanType},
+     * labeled with the provided operation.
+     *
+     * If you need to a span that may complete on another thread, use {@link DetachedSpan#start} instead.
+     */
+    public static CloseableTracer startSpan(Supplier<String> operation, SpanType spanType) {
         Tracer.fastStartSpan(operation, spanType);
         return INSTANCE;
     }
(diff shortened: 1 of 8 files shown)


markelliot:
Instead of continuing to deepen hacks it would be preferable to replace these unfortunate string concatenations with a presentation of the service name where a call takes place in our trace viewing application. That change would easily allow removal of the string munging since it would be obvious from the service boundaries where RPC calls are being made, and thus which side of the interface a particular call lives.

carterkozak:
I agree that would help in some scenarios, however it wouldn't solve the problem where we concatenate the verb and path in [TraceEnrichingFilter](https://github.com/palantir/tracing-java/pull/262/files#diff-648aa4dc0447e11e87cd638a98fc8dabL63) or [Tritium interface names](https://github.com/palantir/tritium/blob/880c6da14f7d89644fce935a5fe0fb8e080b4586/tritium-tracing/src/main/java/com/palantir/tritium/tracing/TracingInvocationEventHandler.java#L70-L72).
You can imagine some of our services are large, and service boundaries aren't sufficient to provide tracing signal, so there tend to be several layers of traced services between endpoint, business logic, and persistence layers.

markelliot:
As far as I recall the monolithic services do not do a ton of string concatenation in their operation descriptions, and I’d continue to assert either way that the string concatenation is likely an issue that should be fixed instead of continuing to jam hacks into this library.

________________________________

carterkozak:
> As far as I recall the monolithic services do not do a ton of string concatenation in their operation descriptions
I wish that were the case.

> I’d continue to assert either way that the string concatenation is likely an issue that should be fixed
How would you fix this in the linked tritium handler?

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -39,6 +40,13 @@ public static CloseableTracer startSpan(String operation) {
         return startSpan(operation, SpanType.LOCAL);
     }
 
+    /**
+     * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
+     */
+    public static CloseableTracer startSpan(Supplier<String> operation) {
```
dansanduleac, line 46: I'm worried that clients might choose to use this instead of the normal one for plain string concatenations which are not that expensive, thereby potentially slowing down hot paths.

carterkozak: Today hot paths are slow depending on the length of the http path.

PR #276, opened 2019-09-12
Propose Tracing2, a smaller, more ergonomic and more efficient tracing

markelliot:
**Note**: Proposing this as a way to discuss future of this library. I think the
implementation is reasonably complete but before proceeding we would want to:
 1. Flesh out correctness tests
 2. Add additional ergonomics examples
 3. Look at what it might take to modify Tracing to become a Tracing2 adapter
 4. Consider what benchmarks might be necessary

## Before this PR
Our API has grown bloated for a number of use cases that were impedence mismatches
with the original library design.

## After this PR
==COMMIT_MSG==
Add Tracing2, a smaller, more ergonomic and more efficient tracing implementation.
==COMMIT_MSG==

## Possible downsides?
Near-term, more API bloat.


carterkozak:
One of the reasons the current API is a mess is that we've coupled the APIs for producers (API that a developer interacts with to instrument a service or library) and consumers (the persistence/logging backend which records sampled data).

For example, there's no reason for us to expose operation names or whether or not an operation is sampled to producers. Leaking that data will result in complex instrumentation. As a goal we want it to be as transparent as possible to add tracing to a library. I would propose modeling this project in a way that it's incredibly difficult to leak these details. We do this in logging by providing slf4j at compile time, and an slf4j implementation as a runtime dependency.

markelliot:
Operation name is an important part of the API. I can take observability out of the API, but don’t believe there’s a real way to decouple it. To take the logging analogy, log producers indicate the level, and so observability as a choice is akin to that. 

I’ll split what I believe is the limited API here, but I also think it’ll result in a pile of service loading that’s grossly unnecessary.

tracing2/src/main/java/com/palantir/tracing2/Spans.java
```diff
@@ -0,0 +1,59 @@
...
+
+public final class Spans {
+    private static final ThreadLocal<Span> currentSpan = new ThreadLocal<>();
+    private static final List<Consumer<CompletedSpan>> consumers = Lists.newCopyOnWriteArrayList();
+
+    public static void clearAllConsumers() {
+        consumers.clear();
+    }
+
+    public static void register(Consumer<CompletedSpan> consumer) {
```
ferozco, line 32: You may want to register consumers by name so that you can also unregister them

tracing2/src/main/java/com/palantir/tracing2/DefaultSpan.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public Span sibling(String op) {
+        close();
+        return create(trace, parent, op);
+    }
+
+    @Override
+    public Span child(String op) {
```
ferozco, line 118: When passing the span across threads its nice to have a `completeAndStartChild` method

ferozco (commented):
I updated the tracing demos and testing lib [here](https://github.com/palantir/tracing-java/pull/296) to get a feel for how easy the new API is to consume

PR #310, opened 2019-10-10
Tracing aware executors wrap Runnables and Callables more cleanly

carterkozak:
==COMMIT_MSG==
Firstly, we should not optimize around stack depth in most cases
because it can result in large methods that the JVM cannot efficiently
inline and additional GC pressure depending on the implementation.

However, we use tracing a lot, a tracing wrapped executor shows up
in ~50% of stack traces in a large internal service, and exceptions
propagate through multiple executors in many cases.
This would not make sense to change if the tracing stack frames
provided information, but wrapping from runnable to callable
isn't helpful, nor are lambda frames to turn a Callable into
a ThrowingCallable.

This change is completely internal to tracing and does not expose
any new public APIs.

Before: frames 2, 3, 4, and 5 are from tracing
```
com.palantir.logsafe.exceptions.SafeRuntimeException: Stack Trace
        at com.palantir.tracing.TracerTest.lambda$testStackDepth$9(TracerTest.java:486)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at com.palantir.tracing.DeferredTracer.withTrace(DeferredTracer.java:104)
        at com.palantir.tracing.Tracers$TracingAwareCallable.call(Tracers.java:356)
        at com.palantir.tracing.WrappingExecutorService.lambda$wrapTask$0(WrappingExecutorService.java:67)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

After: line 2 handles tracing, the rest would be present without tracing
```
com.palantir.logsafe.exceptions.SafeRuntimeException: Stack Trace
        at com.palantir.tracing.TracerTest.lambda$testStackDepth$9(TracerTest.java:486)
        at com.palantir.tracing.Tracers$TracingAwareRunnable.run(Tracers.java:409)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```
==COMMIT_MSG==

## Possible downsides?
none.

### DIFF
--- a/changelog/@unreleased/pr-310.v2.yml
+++ b/changelog/@unreleased/pr-310.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Tracing aware executors wrap Runnables and Callables more cleanly
+  links:
+  - https://github.com/palantir/tracing-java/pull/310
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -16,10 +16,13 @@
 
 package com.palantir.tracing;
 
+import com.google.errorprone.annotations.MustBeClosed;
 import com.palantir.tracing.api.OpenSpan;
 import com.palantir.tracing.api.SpanType;
+import java.io.Closeable;
 import java.io.Serializable;
 import java.util.Optional;
+import java.util.function.Function;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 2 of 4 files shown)


robert3005:
@carterkozak I actually think your description is great as a commit message instead of one sentence summary.

carterkozak:
Good call, updated

iamdanfox:
Is there a way of including an assertion about the stackframes here? The new ones are very neat and don't want to accidentally regress!

carterkozak:
Sure, I did something similar in tritium: https://github.com/palantir/tritium/blob/d69b494360ed69a0739345124028e16e52029821/tritium-lib/src/test/java/com/palantir/tritium/proxy/InstrumentationTest.java#L555-L569

will put it together shortly

carterkozak:
Updated, added docs to the tests to make the intent clear.

iamdanfox (approved):
Thanks Carter!

Issue #316, opened 2019-10-15
It should be possible to remove a span from stack and complete it asynchronously


robert3005:
## What happened?

I am using an interface generated by conjure-java retrofit generator which returns ListenableFuture. What I would like to achieve is being able to parent all work being done to return the listenable future in a span but would like to complete that span only when the future actually completes. Right now I can add a wrapping span that's a sibling to the async spans via DetachedSpan.start and I can add a wrapping span that's a parent that would complete as soon as the synchronous part of the work is done. I believe the issue is that I can only transition from an async to synchronous span (via completeAndStartChild) but I can never convert a synchronous span to async one.

## What did you want to happen?

Not sure what's the best api here but the ask is to effectively be able to detach an existing span and complete it elsewhere. This would let me achieve workflow above by starting a span, making a call, getting result back and detaching a span, registering a callback that closes the span.

carterkozak:
Would something like this work?

```java
DetachedSpan detached = DetachedSpan.start("retrofit-parent");
ListenableFuture<String> result = FluentFuture.from(retrofitClient.greeting())
    .transformAsync(<args>) // any work you need to associate with the future
    .addListener(detached::complete, MoreExecutors.directExecutor());
```

carterkozak:
To make sure I'm following correctly, are you writing this as a conjure-java-runtime feature, or in application code?

robert3005:
so this is library code on top of a conjure retrofit client. The annoying thing with your suggestion is that it's a sibling of all the traces that conjure creates leading to confusing traces

carterkozak:
How about something like this? https://github.com/palantir/tracing-java/pull/323#discussion_r336613094

It avoids the sibling problem, but nests the initial request creation one layer deeper.

robert3005:
#328 provides the necessary functionality - I missed that case before

PR #323, opened 2019-10-18
DetachedSpan#detachThreadLocalSpan

iamdanfox:
## Before this PR

@robert3005 has an async retrofit call that he wants to wrap in a span that he names himself (so that he can differentiate a couple of different possible call sites in trace-log-viewer).

This is currently not possible because the retrofit client reads the trace & span information out of _thread local_ variables, but needs to complete in an async way.

## After this PR
==COMMIT_MSG==
A new `DetachedSpan#detachThreadLocalSpan` method allows converting an in-progress thread local span into a detached span which can be used to instrument async behaviour.
==COMMIT_MSG==

before merging
- [ ] needs example usage (@robert3005?)
- [ ] tests
- [ ] decide if there's a nicer way of supporting this

## Possible downsides?

- @dansanduleac is concerned that code after this invocation will have a different parentSpanId than before, which might not be obvious to users.
- an alternative idea, is to not give users these imperative way of modifying thread local state, and instead just take in a closure?

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
+++ b/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
@@ -49,6 +49,15 @@ public interface DetachedSpan {
     }
 
     /**
+     * Converts the current in-progress thread-local span into a {@link DetachedSpan}, which can be used to more easily
+     * instrument async operations. This {@link DetachedSpan} will need to be manually closed.
+     */
+    @CheckReturnValue
+    static DetachedSpan detachThreadLocalSpan() {
+        return Tracer.detachThreadLocalSpan();
+    }
+
+    /**
      * Equivalent to {@link Tracer#startSpan(String, SpanType)}, but using this {@link DetachedSpan}
      * as the parent instead of thread state.
      */
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -188,6 +188,26 @@ public final class Tracer {
         fastStartSpan(operation, SpanType.LOCAL);
     }
 
+    static DetachedSpan detachThreadLocalSpan() {
+        Trace trace = Preconditions.checkNotNull(
+                currentTrace.get(),
+                "detachThreadLocalSpan requires an in-progress trace with a thread-local span");
+
+        // pop current thread-local span, so that it won't be used as parent for subsequent spans
+        Optional<OpenSpan> span = trace.pop();
+
+        // to maintain the invariant that thread local traces always have at least one span in progress we make up a
+        // span to leave on this thread
+        if (trace.isEmpty()) {
+            String spanLeftBehind = trace.isObservable() ? span.get().getOperation() + " (left behind)" : "";
+            trace.fastStartSpan(spanLeftBehind, SpanType.LOCAL);
+        }
+
+        return trace.isObservable()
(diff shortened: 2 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -48,6 +48,15 @@ static DetachedSpan start(String operation, SpanType type) {
...
     }
 
+    /**
+     * Converts the current in-progress thread-local span into a {@link DetachedSpan}, which can be used to more easily
+     * instrument async operations. This {@link DetachedSpan} will need to be manually closed.
+     */
+    @CheckReturnValue
+    static DetachedSpan detachThreadLocalSpan() {
+        return Tracer.detachThreadLocalSpan();
+    }
```
carterkozak, line 58: I think this is a foot-gun. It's easy to refactor the start/complete attached span pieces without remembering to remove an detachThreadLocalSpan or migrate it to  create a new detached span.

robert3005:
@iamdanfox you can't take a closure since at no point we complete the detached span, see #316 for the explanation of the issue

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
robert3005: @carterkozak would love to know what you suggest here. #316 is the issue we were trying to solve.

carterkozak: A couple potential options (neither of which is ideal):
1. We create child spans for subsequent operations. The output may not be pretty, but it will give us more information about the processing flow.
2. Create a single detached span for the duration of the call. It's not attached to a thread, so logging will not be associated with the traceId, but we get pretty traces without risking clobbering thread state.

Is the delta between existing options and the ideal worth the risk of corrupted tracing?

robert3005: Let me explain a little bit more what I'd like to do so we are on the same page.

I have a function that does something like

```
ListenableFuture<A> valueA = callA();
ListenableFuture<B> valueB = callB();
ListenableFuture<C> valueC = callC();
```

All of those functions `callA`, `callB`, `callC` do some work synchronously and produce traces. I would like to have all calls to those functions have distinct so imagine I would wrap all those functions calls in something like

```
Tracer.fastStartSpan("call to A");
ListenableFuture<A> valueA = callA();
DetachedSpan span = Tracer.detachThreadLocalSpan();
Futures.addCallback(result, new FutureCallback<>() {
    @Override
    public void onSuccess(@Nullable TemporaryCredentialsAuthToken result) {
        span.complete();
    }

    @Override
    public void onFailure(Throwable t) {
        span.complete();
    }
}, MoreExecutors.directExecutor());
return result;
```
After thinking about how to deal with this I couldn't figure out how to do it without detach span. 

Creating a detached span for duration of the call is confusing since you don't know parent relations and you only use the start times as a proxy which leads to confusing traces

carterkozak: Does this work?

```java
DetachedSpan detached = DetachedSpan.start("call to A");
ListenableFuture<A> valueA;
try (CloseableSpan span = detached.childSpan("Initiate call to A")) {
    valueA = callA();
}
valueA.addListener(detached::complete, MoreExecutors.directExecutor());
```

robert3005: I thought there was something weird about this one but let me try again

PR #328, opened 2019-10-23
Tracers.wrapListenableFuture utility to wrap ListenableFutures with detached spans

carterkozak:
## Before this PR
Required one-off implementation for each consumer, with potential for confusing output and bugs. Here we have a documented utility with test coverage.

## After this PR
==COMMIT_MSG==
Tracers.wrapListenableFuture utility to wrap ListenableFutures with detached spans
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -202,7 +202,7 @@ public final class Tracers {
         // no active trace, the detached span would not be associated with work initiated by delegateFactory.
         try (CloseableSpan ignored =
                 // This could be more efficient using https://github.com/palantir/tracing-java/issues/177
-                span.childSpan("Initial: " + operation)) {
+                span.childSpan(operation + " initial")) {
             result = Preconditions.checkNotNull(delegateFactory.get(), "Expected a ListenableFuture");
         } finally {
             if (result != null) {
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -16,11 +16,16 @@
 
 package com.palantir.tracing;
 
+import com.google.common.annotations.Beta;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.logsafe.Preconditions;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
 
 /** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Tracers {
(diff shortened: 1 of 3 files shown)


tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -174,6 +179,45 @@ public static Runnable wrap(String operation, Runnable delegate) {
...
+    @Beta
+    public static <T, U extends ListenableFuture<T>> U wrapListenableFuture(
+            String operation, Supplier<U> delegateFactory) {
+        DetachedSpan span = DetachedSpan.start(operation);
+        U result = null;
+        // n.b. This span is required to apply tracing thread state to an initial request. Otherwise if there is
+        // no active trace, the detached span would not be associated with work initiated by delegateFactory.
+        try (CloseableSpan ignored =
+                // This could be more efficient using https://github.com/palantir/tracing-java/issues/177
+                span.childSpan("Initial: " + operation)) {
```
iamdanfox: I think I'd like to keep the user-supplied operation name at the beginning so that it's easier to scan in UIs (especially when the span name appears in a column which might be too narrow).

Thoughts on just appending ` (initial)` or ` (construct ListenableFuture)` ?

iamdanfox:
@robert3005 I think this would let you pick a name for your span and keep all the implementation details parented inside it:

image

robert3005:
I believe so. Somehow I convinced myself this wouldn't work earlier. I should have traces from production shortly

tracing/src/main/java/com/palantir/tracing/Tracers.java
carterkozak: sounds reasonable to me.

robert3005:
LGTM, thanks @carterkozak

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -257,6 +264,120 @@ public void testWrapRunnable_startsNewSpan() throws Exception {
         assertThat(Tracer.completeSpan().get().getOperation()).isEqualTo("outside");
     }
 
+    @Test
+    public void testWrapFuture_immediate() {
+        List<Span> observed = new ArrayList<>();
+        Tracer.subscribe("futureTest", observed::add);
```
iamdanfox, line 270: shall we just make this a TestSubscriber? it would be a teeny class but might make all of our unit-tests a little shorter!

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -257,6 +264,120 @@ public void testWrapRunnable_startsNewSpan() throws Exception {
...
+                    .extracting(Span::getOperation)
+                    .isEqualTo(operationName);
+            assertThat(observed)
+                    .allSatisfy(span -> assertThat(span)
+                            .extracting(Span::getTraceId)
+                            .isEqualTo("defaultTraceId"));
+        } finally {
+            Tracer.unsubscribe("futureTest");
+        }
+    }
```
iamdanfox, line 291: If we'd fully converted this repo to JUnit5 then I would have loved to use the @TestTracing things for these!

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -257,6 +264,120 @@ public void testWrapRunnable_startsNewSpan() throws Exception {
...
```
iamdanfox, line 317: worth having a test where we cancel the future too?

iamdanfox (approved):

tracing/src/test/java/com/palantir/tracing/TracersTest.java
carterkozak: I could go either way, cancel isn't any different from exceptional completion which is already tested.

robert3005:
fixes #316

PR #343, opened 2019-11-07
Add methods to create CloseableTracer with metadata

pkoenig10:
## Before this PR
It was not possible to use `CloseableTracer` and close the span with metadata.

## After this PR
It is possible to use `CloseableTracer` and close the span with metadata.


carterkozak:
Why is this necessary? As far as I'm aware, none of our implementations are capable of persisting metadata, it should probably be deprecated and removed.

pkoenig10:
The metadata isn't necessary. I'm happy to stop using it if we have no plans to support persisting metadata at any point in the future.

If this is the case, then we should deprecate the relevant methods to discourage the use of metadata.

PR #347, opened 2019-11-11
Deprecate methods to complete spans with metadata

pkoenig10:
See https://github.com/palantir/tracing-java/pull/343#issuecomment-551288972.


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -373,8 +376,11 @@ private static void completeSpanAndNotifyObservers(
     /**
      * Like {@link #completeSpan()}, but adds {@code metadata} to the current span being completed.
      * If the return value is not used, prefer {@link Tracer#fastCompleteSpan(Map)}.
+     *
+     * @deprecated Use {@link #completeSpan()}
```
```suggestion
     * @deprecated Use {@link #fastCompleteSpan()}
```

carterkozak (approved):

carterkozak:
Sorry for the delay, this fell off my radar

PR #350, opened 2019-11-14
Replace Trace headers instead of append

### LINKED ISSUE #349: OkHttpTraceInterceptor should replace trace headers, rather than append
## What happened?

Currently traces are appended, resulting in invalid trace headers e.g `51a5a6fad8b26bcd,51a5a6fad8b26bcd`

Trace header should be replaced, to avoid multi-value headers. This can be achieved using `Request.Builder#header` instead of `Request.Builder#addHeader`

gatesn:
Fixes #349 

As per https://github.com/openzipkin/b3-propagation, the trace headers shouldn't contain multiple values.

### DIFF
--- a/changelog/@unreleased/pr-350.v2.yml
+++ b/changelog/@unreleased/pr-350.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Replace Trace headers instead of appending them.
+  links:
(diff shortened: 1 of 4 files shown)


iamdanfox:
Oof that's kinda trippy, thanks @ngates!

iamdanfox (approved):

PR #352, opened 2019-11-14
Introduce topSpanId to disambiguate incoming requests that share traceid

JacekLach:
## Before this PR

Logs from concurrent incoming requests with the same traceid are currently difficult
to tell apart (one can look at thread names etc, but not across executors).

## After this PR
==COMMIT_MSG==
This introduces a new optional property on a Trace, topSpanId, that can be read to disambiguate span stacks that belong to the same traceid bug had different entry points.

Use that in tracing-undertow and tracing-jersey to tell apart requests with the same traceid.
==COMMIT_MSG==

## Possible downsides?

The unsampled span has to keep a bit more state (not just the parent of the top span as per originating span id, but also the top span id).
We have to generate a new span id in unsampled spans if the stack is empty.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -29,6 +29,7 @@ import com.palantir.tracing.api.SpanType;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.Optional;
+import javax.annotation.Nullable;
 
 /**
  * Represents a trace as an ordered list of non-completed spans. Supports adding and removing of spans. This class is
@@ -44,10 +45,13 @@ import java.util.Optional;
 public abstract class Trace {
 
     private final String traceId;
+    @Nullable
+    private final String localTraceId;
 
-    private Trace(String traceId) {
+    private Trace(String traceId, @Nullable  String localTraceId) {
         checkArgument(!Strings.isNullOrEmpty(traceId), "traceId must be non-empty");
         this.traceId = traceId;
+        this.localTraceId = localTraceId;
     }
 
     /**
@@ -136,27 +140,44 @@ public abstract class Trace {
         return traceId;
     }
 
+    /**
+     * The globally unique non-empty identifier for this call trace within a service (or another locality context).
+     *
+     * While {@link #getTraceId()} is expected to be propagated across RPC calls, localTraceId distinguishes between
+     * two concurrent RPC calls made to a service with the same traceid.
+     */
+    final Optional<String> getLocalTraceId() {
+        // XXX: should this be equal to traceId if localTraceId is not set?
+        return Optional.ofNullable(localTraceId);
+    }
+
     abstract Optional<String> getOriginatingSpanId();
 
     /** Returns a copy of this Trace which can be independently mutated. */
     abstract Trace deepCopy();
 
     static Trace of(boolean isObservable, String traceId) {
-        return isObservable ? new Sampled(traceId) : new Unsampled(traceId);
+        return isObservable ? new Sampled(traceId, null) : new Unsampled(traceId, null);
+    }
+
+    static Trace of(boolean isObservable, String traceId, String localTraceId) {
+        checkArgument(!Strings.isNullOrEmpty(localTraceId), "localTraceId must be non-empty");
+        return isObservable ? new Sampled(traceId, localTraceId) : new Unsampled(traceId, localTraceId);
     }
 
     private static final class Sampled extends Trace {
 
         private final Deque<OpenSpan> stack;
 
-        private Sampled(ArrayDeque<OpenSpan> stack, String traceId) {
-            super(traceId);
+        private Sampled(ArrayDeque<OpenSpan> stack, String traceId, @Nullable String localTraceId) {
+            super(traceId, localTraceId);
             this.stack = stack;
         }
 
-        private Sampled(String traceId) {
-            this(new ArrayDeque<>(), traceId);
+        private Sampled(String traceId, @Nullable String localTraceId) {
+            this(new ArrayDeque<>(), traceId, localTraceId);
         }
+
 
         @Override
         @SuppressWarnings("ResultOfMethodCallIgnored") // Sampled traces cannot optimize this path
@@ -205,7 +226,7 @@ public abstract class Trace {
 
         @Override
         Trace deepCopy() {
-            return new Sampled(new ArrayDeque<>(stack), getTraceId());
+            return new Sampled(new ArrayDeque<>(stack), getTraceId(), getLocalTraceId().orElse(null));
         }
 
         @Override
@@ -222,14 +243,14 @@ public abstract class Trace {
         private int numberOfSpans;
         private Optional<String> originatingSpanId = Optional.empty();
 
-        private Unsampled(int numberOfSpans, String traceId) {
-            super(traceId);
+        private Unsampled(int numberOfSpans, String traceId, @Nullable String localTraceId) {
+            super(traceId, localTraceId);
             this.numberOfSpans = numberOfSpans;
             validateNumberOfSpans();
         }
 
-        private Unsampled(String traceId) {
-            this(0, traceId);
+        private Unsampled(String traceId, @Nullable String localTraceId) {
+            this(0, traceId, localTraceId);
         }
 
         @Override
@@ -289,7 +310,7 @@ public abstract class Trace {
 
         @Override
         Trace deepCopy() {
-            return new Unsampled(numberOfSpans, getTraceId());
+            return new Unsampled(numberOfSpans, getTraceId(), getLocalTraceId().orElse(null));
         }
 
         /** Internal validation, this should never fail because {@link #pop()} only decrements positive values. */
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -236,17 +236,23 @@ public abstract class Trace {
          * This allows thread trace state to be cleared when all "started" spans have been "removed".
          */
         private int numberOfSpans;
-        private Optional<String> originatingSpanId = Optional.empty();
-        private Optional<String> topSpanId = Optional.empty();
+        private Optional<String> originatingSpanId;
+        private Optional<String> topSpanId;
 
-        private Unsampled(int numberOfSpans, String traceId) {
+        private Unsampled(
+                int numberOfSpans,
+                String traceId,
+                Optional<String> originatingSpanId,
+                Optional<String> topSpanId) {
             super(traceId);
             this.numberOfSpans = numberOfSpans;
+            this.originatingSpanId = originatingSpanId;
+            this.topSpanId = topSpanId;
             validateNumberOfSpans();
         }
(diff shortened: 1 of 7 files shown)


JacekLach:
Followups are to include this new property in request and service logs

iamdanfox:
I am not super comfortable about diverging from the concepts that zipkin already define: https://zipkin.io/pages/architecture.html.. we've already got james' `X-B3-Originating-Span-Id` which we already put on the wire: https://github.com/palantir/tracing-java/blob/2150bd2886d36adb9ba078de350d1d31f50f8ba9/tracing-okhttp3/src/test/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptorTest.java#L75-L78

~if normal service logs had james' originating span id as a param somehow, wouldn't you be able to achieve the desired outcome (of differentiating different concurrent requests to one service from the same caller even if they aren't sampled?)~

Edit nvm seems like these would be the same

JacekLach:
Yeah; `ORIGINATING_SPAN_ID` relates together requests made from the same caller within a traceid (and disambiguates them from requests hitting current service from a different caller, within the traceid).

However what we're trying to do here is tell apart two requests from the same caller, so we need to generate this id within the current service. (or rely on caller to never call us with the same `x-b3-spanid`)

JacekLach:
`tracing-java` is pretty much the only place we can do this in, because we want all the usual trace propagation (executors, thread changes, etc) to apply to this property.

carterkozak:
I'm not sure this is quite what we want, in the past we have discussed service-local request identifiers, which we can use to associate request data with all service logging for that request, not necessarily connected to tracing, though there are definitely parallels.
That way we could build queries for service logging from product `Foo` for requests to endpoint `GET /api/bar`.

JacekLach:
Well, if we don't connect this to tracing then it has to be propagated manually when you say submit something to an executor while processing a request, right?

I don't think there's a point to having multiple thread-locals that we carry around to mark 'logically still the same unit of work'. Or is that not what you meant?

> That way we could build queries for service logging from product Foo for requests to endpoint GET /api/bar.

This seems like a different question - you want to see what the request path for a service log line was? Sure, seems possibly useful, but unrelated to my problem :P

carterkozak:
> unrelated to my problem :P

Wouldn't it allow us to associate a span ID to the request, and the request ID to all request logging, solving your problem?

JacekLach:
That assumes that the incoming span id is unique, which is the case for conjure clients currently but that seems like an implementation detail.
(unless you mean we'd generate a fresh id anyway? but then that seems to reduce to this pr, just bigger)

sfackler:
The server generates a new span for the requests parented to the span sent by the client, so that one will always be unique.

JacekLach:
>The server generates a new span for the requests parented to the span sent by the client, so that one will always be unique.

Yeah, but the Trace can't really know that the second-to-oldest span in the stack is the special local-unique one.

Also, it's totally possible to close the request serving span (return a value) but still want to log things related to that request with the same traceid / localTraceId. For example you might be submitting an asynchronous cleanup task to run after your request completes, which is related to the trace/request, but not contained within it.

sfackler:
You can parent a span to one that's already closed.

JacekLach:
Hm, actually, I thought that the originating span id was the id of the top span in the stack, but actually it's the *parent* of the top span in the stack. So this can be done without the extra field / constructors. Will change

JacekLach:
e57846e - we no longer maintain a separate id for the local trace, and instead expose the id of the topmost span in the stack.

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -234,22 +253,27 @@ private Unsampled(String traceId) {
...
         @Override
         protected void push(OpenSpan span) {
-            startSpan(span.getParentSpanId());
-        }
-
-        private void startSpan(Optional<String> parentSpanId) {
             if (numberOfSpans == 0) {
-                originatingSpanId = parentSpanId;
+                // TODO: shouldn't this be span.getOriginatingSpanId?
+                originatingSpanId = span.getParentSpanId();
```
JacekLach, line 278: When we push an existing span as the root span of a trace, shouldn't we preserve the originating span id?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -282,13 +307,19 @@ boolean isObservable() {
...
+
         @Override
         Optional<String> getOriginatingSpanId() {
             return originatingSpanId;
         }
 
         @Override
         Trace deepCopy() {
+            // TODO: shouldn't this preserve originatingSpanId / topSpanId?
             return new Unsampled(numberOfSpans, getTraceId());
```
JacekLach: Copying an `Unsampled` trace discards originating / top span ids. (a sampled one does not). This seems like a bug?

JacekLach:
(check failures are just the TODOs, which I'm planning to remove when I know if the current behaviour is a bug or not)

tracing/src/main/java/com/palantir/tracing/Trace.java
JacekLach: Chatted with @dansanduleac - we think this means requests made in deferred traces won't have an originating span id header set

tracing/src/main/java/com/palantir/tracing/Trace.java
dansanduleac: Yep i think you're correct.
However, because of the `if` above, this code would only ever be hit when the `span.getParentSpanId()` is the same as `span.getOriginatingSpanId()`, because `Trace.getOriginatingSpanId()` would always be empty when there are no spans.

dansanduleac: Honestly I don't even know why OpenSpan keeps a copy of `getOriginatingSpanId` because the Trace also stores it and should be the authority, given this number should be unique for a given Trace.
I think if we propagated this correctly in `Trace.deepCopy` then we could get rid of it from `OpenSpan` and this sort of question you're legitimately asking would become trivial to answer.

tracing/src/main/java/com/palantir/tracing/Trace.java
dansanduleac: Actually looked at the code and there is something broken here indeed, but not where you think.

DeferredTrace actually copies twice: the first time in its constructor, it's only as a hack, in order to get access to the Trace, but the 2nd time it's to save (and then restore) the original trace.
It's this 2nd time where interesting things can happen.

Note that DeferredTrace keeps track of the `parentSpanId` and the inner "deferred" Trace it creates is actually empty, as in you don't carry around the existing span stack. When you call `.withTrace` to use the deferred trace, it then calls `fastStartSpan` with that saved `parentSpanId`, thus ensuring that the new Trace initializes its `originatingSpanId` correctly from that.
All good so far.

Now it gets tricky when restoring the original trace.
We create a new Trace, losing the information as you correctly point out.
For Sampled, *new* spans are able to recover the correct `originatingSpanId` because `Sampled.getOriginatingSpanId()` grabs it from the last item in the stack (i.e. the top).
For *Unsampled* though, `getOriginatingSpanId` will now be cleared and unable to recover the original value. 🌶🌶🌶

dansanduleac:
All in all I think we should wire up the propagating the originatingSpanId (and your topSpanId) through the Trace, and remove the originatingSpanId from the OpenSpan as it doesn't belong there (as I explained in more detail in https://github.com/palantir/tracing-java/pull/352#discussion_r347529258)

tracing/src/main/java/com/palantir/tracing/Trace.java
JacekLach: Ah, I didn't realise `push` was only called from `startSpan` (in which case yeah we can assert that `span.getOriginatingSpanId()` is equal to `span.getParentId()` if we're empty)

tracing/src/main/java/com/palantir/tracing/Trace.java
JacekLach: >Note that DeferredTrace keeps track of the parentSpanId and the inner "deferred" Trace it creates is actually empty, as in you don't carry around the existing span stack.

Oh so curiously the originating span id of a deferred trace is the innermost span that created the deferred trace, not the originating span id of the trace at that point?

Not sure if this is intentional.

I guess we don't copy the entire stack because we expect those spans to get closed before the deferred trace actually gets to run?

dansanduleac: I think you are right... that doesn't sound right to me either.

I think we don't copy the entire stack because we want to enforce those spans to not be involved in the deferred span trace, i.e. you are not allowed to close them.

dansanduleac:
I just noticed this, but since `Trace.top()` already exists and gets you the innermost span, which is the opposite of what you're introducing, we should probably call your topSpanId something else - maybe `outermostSpanId`?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -220,36 +236,47 @@ public String toString() {
...
-            this(0, traceId);
+            this(0, traceId, Optional.empty(), Optional.empty());
         }
 
         @Override
         void fastStartSpan(String _operation, String parentSpanId, SpanType _type) {
-            startSpan(Optional.of(parentSpanId));
+            if (numberOfSpans == 0) {
+                originatingSpanId = Optional.of(parentSpanId);
+                topSpanId = Optional.of(Tracers.randomId());
```
carterkozak: Do we ever want to generate a new random ID if the `SpanType` is not `SERVER_INCOMING`? In all other cases I think the traceId we already have should be sufficient.

carterkozak (commented):
It doesn't look like this uses the top-span identifier. If we implement this approach, how do we take advantage of the value without requiring all products to make code changes?

JacekLach:
>It doesn't look like this uses the top-span identifier. If we implement this approach, how do we take advantage of the value without requiring all products to make code changes?

Followup PR is to update sls-logging to inject the identifier as an extra
param to service logs. (that's what the jersey interceptor change
facilitates)

No opinion on filtering to only the SERVER_INCOMING span types; I assume
CLIENT_OUTGOING spans should ~never be the root span, so the question is
whether this might be useful for local spans? Which would require someone
starting multiple span stacks with the same traceid.

carterkozak:
Notes from a sync with @JacekLach @iamdanfox @dansanduleac and myself, tagging @ferozco for context as we chatted previously:

This is a divergence from standard zipkin, so we should be careful that we don't implement it in a way that prevents us from future features or refactors. However, tracing state is correctly propagated through most applications, and adding additional state would require a heavy investment. In particular utilities for asynchronous tracing (executors, detached traces, etc) would need to be duplicated across every codebase to additionally handle requestId, which is clearly not desirable.

We should implement this without any change to public API, this way we're not locked into supporting request identifiers in the tracing library for perpetuity, only that we maintain the functionality to associate logging with a request.

We can achieve this by setting a known MDC key, `_requestId`, the same way we set `traceId` and `_sampled`. This value should be initialized only when a new **root** span is created with span type `SERVER_INCOMING`.

iamdanfox:
linking to the in-progress PR https://github.com/palantir/tracing-java/pull/364

PR #362, opened 2019-11-22
Scheduled executor recurring tasks are always given new trace IDs

carterkozak:
==COMMIT_MSG==
Previously we had no good options for scheduled executors: Either
create a new traceId for every taskd regardless of whether it was
submitted to execute(runnable) or scheduleWithFixedDelay which
mean very different things.

Recurring tasks are most frequently scheduled on application startup
and continue to execute at a cadence until the service is halted.
In these cases we used to produce large traces bounded by the
runtime of the application, making the traceId of an error thrown
in this context unhelpful for debugging. In worse cases, lazily
scheduled tasks are created with the traceId of a request that
happened to cause lazy initialization, resulting in tracing state
that incorrectly blames to a particular user request.

While there are cases that we may want to preserve a single traceId
across a scheduled task, they are sparse and complex. In these cases
it's reasonable to expect developers to wrap tasks before scheduling
them with an executor to acknowledge the desired behavior, and make
it clear to others who read the code.
==COMMIT_MSG==

## Possible downsides?
Risk for cases where a single traceId is expected across recurring tasks.
This is justifiable because there are very few of these (none that I am
personally aware of) and many cases that are fixed by this change.

While there are cases that we may want to preserve a single traceId
across a scheduled task, they are sparse and complex. In these cases
it's reasonable to expect developers to wrap tasks before scheduling
them with an executor to acknowledge the desired behavior, and make
it clear to others who read the code.

### DIFF
--- a/changelog/@unreleased/pr-362.v2.yml
+++ b/changelog/@unreleased/pr-362.v2.yml
@@ -1,0 +1,26 @@
+type: break
+break:
+  description: |-
+    Scheduled executor recurring tasks are always given new trace IDs
+
+    Previously we had no good options for scheduled executors: Either
+    create a new traceId for every taskd regardless of whether it was
+    submitted to execute(runnable) or scheduleWithFixedDelay which
+    mean very different things.
+
+    Recurring tasks are most frequently scheduled on application startup
+    and continue to execute at a cadence until the service is halted.
+    In these cases we used to produce large traces bounded by the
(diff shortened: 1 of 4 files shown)


carterkozak:
I'm going to err on the side of caution and call this a break.

iamdanfox (approved):
Embarrassingly I thought this was already the case - definitely in sync this is desirable behaviour!

PR #364, opened 2019-11-25
Introduce requestId MDC property

JacekLach:
Logs from concurrent incoming requests with the same traceid are currently difficult
to tell apart (one can look at thread names etc, but not across executors).

## After this PR
==COMMIT_MSG==
Introduce a new MDC property managed by Tracers, which stores a unique identifier
when the top span of the current trace is a SERVER_INCOMING span.
==COMMIT_MSG==

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
@@ -29,6 +29,7 @@ import io.undertow.server.HttpServerExchange;
 import io.undertow.util.AttachmentKey;
 import io.undertow.util.HeaderMap;
 import io.undertow.util.HttpString;
+import org.slf4j.MDC;
 
 /**
  * Extracts Zipkin-style trace information from the given HTTP request and sets up a corresponding
@@ -61,32 +62,12 @@ public final class TracedOperationHandler implements HttpHandler {
 
     @Override
     public void handleRequest(HttpServerExchange exchange) throws Exception {
-        HeaderMap headers = exchange.getRequestHeaders();
-        // TODO(rfink): Log/warn if we find multiple headers?
-        String traceIdHeader = headers.getFirst(TRACE_ID); // nullable
-        String traceId;
-        // XXX: again, nasty abstraction leak
-        String requestId;
-
-        // Set up thread-local span that inherits state from HTTP headers
-        if (Strings.isNullOrEmpty(traceIdHeader)) {
-            traceId = initializeNewTrace(headers);
-            requestId = Tracer.startSpan(operation, SpanType.SERVER_INCOMING).getSpanId();
-        } else {
-            traceId = traceIdHeader;
-            Tracer.initTrace(getObservabilityFromHeader(headers), traceIdHeader);
-            String spanId = headers.getFirst(SPAN_ID); // nullable
-            if (spanId == null) {
-                requestId = Tracer.startSpan(operation, SpanType.SERVER_INCOMING).getSpanId();
-            } else {
-                requestId = Tracer.startSpan(operation, spanId, SpanType.SERVER_INCOMING).getSpanId();
-            }
-        }
+        String traceId = initializeTrace(exchange);
 
         // Populate response before calling delegate since delegate might commit the response.
         exchange.getResponseHeaders().put(TRACE_ID, traceId);
         exchange.putAttachment(IS_SAMPLED_ATTACHMENT, Tracer.isTraceObservable());
-        exchange.putAttachment(REQUEST_ID_ATTACHMENT, requestId);
+        exchange.putAttachment(REQUEST_ID_ATTACHMENT, MDC.get(Tracers.REQUEST_ID_KEY));
         try {
             delegate.handleRequest(exchange);
         } finally {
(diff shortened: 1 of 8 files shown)


tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
```diff
@@ -60,11 +61,32 @@ public TracedOperationHandler(HttpHandler delegate, String operation) {
...
-        String traceId = initializeTrace(exchange);
+        HeaderMap headers = exchange.getRequestHeaders();
+        // TODO(rfink): Log/warn if we find multiple headers?
+        String traceIdHeader = headers.getFirst(TRACE_ID); // nullable
+        String traceId;
+        // XXX: again, nasty abstraction leak
+        String requestId;
+
+        // Set up thread-local span that inherits state from HTTP headers
+        if (Strings.isNullOrEmpty(traceIdHeader)) {
```
JacekLach: I had to inline the trace creation to get both the span id and the trace id out, without creating a wrapper object. Maybe just reading the MDC here is simpler?

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -136,6 +136,8 @@ final String getTraceId() {
         return traceId;
     }
 
+    abstract Optional<String> getRequestId();
```
JacekLach, line 139: We still store the request id so that it can be restored when a trace is set. However this getter is not exposed to clients of `Tracers` / `Tracer`

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -146,7 +146,13 @@ public static void initTrace(Observability observability, String traceId) {
      */
     @CheckReturnValue
     public static OpenSpan startSpan(String operation, String parentSpanId, SpanType type) {
-        return getOrCreateCurrentTrace().startSpan(operation, parentSpanId, type);
+        Trace trace = getOrCreateCurrentTrace();
+        boolean wasEmpty = trace.isEmpty();
```
JacekLach, line 150: This is kinda sad - but definitely don't want to be setting the MDC property on every span open, so we have to know if it could have become active here. If traces had to be initialized with a span, this would be part of `init/setTrace` :(

tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedOperationHandler.java
carterkozak: Ya, I'd prefer the minimal change where we read the requestId from MDC here.

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -195,6 +197,18 @@ boolean isObservable() {
...
+        @Override
+        Optional<String> getRequestId() {
+            if (stack.isEmpty()) {
+                return Optional.empty();
+            }
+            OpenSpan top = stack.peekFirst();
+            if (top.type() != SpanType.SERVER_INCOMING) {
+                return Optional.empty();
+            }
+            return Optional.of(top.getSpanId());
```
carterkozak, line 209: I think we have to pass around the value like originatingSpanId, when we create deferred traces we don't pass the full stack of spans (we don't want to allow other code to complete spans from our thread) but we still want to pass the same requestId.

carterkozak:
Perhaps it's time to deprecate `Tracer.initTrace` in favor of a method to initialize a trace with a root span. That way we can continue to support the deprecated method for now (without this new functinoality), and support the new, safe things with the new method.

tracing/src/main/java/com/palantir/tracing/Trace.java
JacekLach: Hm, but we don't do that for `originatingSpanId`, do we? Not for sampled spans. So while I agree with this, I think it should be fixed for originatingSpanId and requestId together in a follow-up pr

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -155,7 +161,13 @@ public static OpenSpan startSpan(String operation, String parentSpanId, SpanType
      */
     @CheckReturnValue
     public static OpenSpan startSpan(String operation, SpanType type) {
-        return getOrCreateCurrentTrace().startSpan(operation, type);
+        Trace trace = getOrCreateCurrentTrace();
+        boolean wasEmpty = trace.isEmpty();
+        OpenSpan result = trace.startSpan(operation, type);
+        if (wasEmpty) {
+            setRequestIdIfPresent(trace.getRequestId());
```
dansanduleac, line 168: This feels like the wrong place to do this. 
I think we should do what carter recommended (https://github.com/palantir/tracing-java/pull/364#issuecomment-558700652) which then means that this only has to happen in one code path.

dansanduleac (commented):
I think we should instead
1. initialize the requestId once via a new API (as Carter recommended), when creating a new trace initialized via a SERVER_INCOMING
2. set the MDC when _setting_ a trace as the current trace (i.e. in `Tracer.setTrace`, just like we set the traceId MDC key) which has the advantage that whenever we apply a deferred trace on another thread (via `DeferredTracer.withTrace`) it will set the `requestId` at the right time too

@carterkozak thoughts? I had initial context on this but it seems like you/@ferozco have been review-requested by the load-balancing bots

carterkozak:
Yep, that's exactly what I was thinking, thanks Dan.

carterkozak:
Proposal is up #574

carterkozak:
Closing this out because #574 has been released.

PR #375, opened 2019-12-07
New endpoint to wrap Callable with alternative trace ID

jiahuijiang:
## Before this PR

Currently there is no endpoint to wrap callables with provided trace ID, only for runnable

## After this PR
A new endpoint to support wrapping Callables with alternative trace ID is added

## Possible downsides?

There can be issue with type inference with existing usage of wrapWithAlternativeTraceId -> if the current passed in runnable has returned value, then now it will be inferred as Callable and can be treated as a break

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -334,30 +334,6 @@ public final class Tracers {
     }
 
     /**
-     * Wraps the given {@link Callable} such that it creates a fresh {@link Trace tracing state with the given traceId}
-     * for its execution. That is, the trace during its {@link Callable#call() execution} will use the traceId provided
-     * instead of any trace already set on the thread used to execute the callable. Each execution of the callable
-     * will use a new {@link Trace tracing state} with the same given traceId. The given {@link String operation} is
-     * used to create the initial span.
-     */
-    public static <V> Callable<V> wrapWithAlternateTraceId(String traceId, String operation,
-            Observability observability, Callable<V> delegate) {
-        return () -> {
-            // clear the existing trace and keep it around for restoration when we're done
-            Optional<Trace> originalTrace = Tracer.getAndClearTraceIfPresent();
-
-            try {
-                Tracer.initTrace(observability, traceId);
-                Tracer.fastStartSpan(operation);
-                return delegate.call();
-            } finally {
-                Tracer.fastCompleteSpan();
-                restoreTrace(originalTrace);
-            }
-        };
-    }
-
-    /**
      * Deprecated.
      *
      * @deprecated Use {@link #wrapWithNewTrace(String, Runnable)}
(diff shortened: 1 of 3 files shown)


tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -382,7 +406,7 @@ public static Runnable wrapWithAlternateTraceId(String traceId, Runnable delegat
     public static Runnable wrapWithAlternateTraceId(String traceId, String operation, Runnable delegate) {
         return wrapWithAlternateTraceId(traceId, operation, Observability.UNDECIDED, delegate);
     }
-
+/*
```
dansanduleac: hmm

jiahuijiang: @dansanduleac we do have a copy of all endpoints one for Runnable one for Callable. I don't see why wrapWithAlternateTraceId shouldn't :-P

dansanduleac: i mean, this line should be empty no?

jiahuijiang: 🤦‍♂

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -369,6 +369,30 @@ public static Runnable wrapWithNewTrace(String operation, Observability observab
...
 
+    /**
+     * Wraps the given {@link Callable} such that it creates a fresh {@link Trace tracing state with the given traceId}
+     * for its execution. That is, the trace during its {@link Callable#call() execution} will use the traceId provided
+     * instead of any trace already set on the thread used to execute the callable. Each execution of the callable
+     * will use a new {@link Trace tracing state} with the same given traceId. The given {@link String operation} is
+     * used to create the initial span.
+     */
+    public static <V> Callable<V> wrapWithAlternateTraceId(String traceId, String operation,
+            Observability observability, Callable<V> delegate) {
```
dansanduleac, line 380: This could introduce an abiguity at the compilation level between this and the override which takes a Runnable, but I've verified that there are only 3 invocations internally and so we should be fine.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -522,6 +522,101 @@ public void testWrapCallableWithNewTrace_canSpecifyObservability() throws Except
...
+        assertThat(traceIdInsideCallable)
+                .isNotEqualTo(traceIdBeforeConstruction)
+                .isNotEqualTo(traceIdAfterCall)
+                .isEqualTo(traceIdToUse);
+
+        assertThat(traceIdBeforeConstruction).isEqualTo(traceIdAfterCall);
+    }
+
+    @Test
+    public void testWrapCallableWithAlternateTraceId_traceStateInsideRunnableHasSpan() throws Exception {
```
dansanduleac:
```suggestion
    public void testWrapCallableWithAlternateTraceId_traceStateInsideCallableHasSpan() throws Exception {
```

dansanduleac:
👍

PR #387, opened 2020-01-07
Improvement: Make Tracer#getTraceMetadata public

robert3005:
## Before this PR
It's hard/impossible to integrate tracing with non okhttp http libraries as there's no way to access the same metadata as okhttptraceinterceptor has access to.

## After this PR
==COMMIT_MSG==
Make Tracer#getTraceMetadata public to allow third party libraries to propagate trace related ids
==COMMIT_MSG==

## Possible downsides?
We add yet another thing to the api surface

cc @markelliot

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -97,7 +97,6 @@ public final class Tracer {
      * <p>
      * n.b. this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
      */
-    @Beta
     public static TraceMetadata getTraceMetadata() {
         Trace trace = checkNotNull(currentTrace.get(), "Unable to getTraceMetadata when there is trace in progress");
 
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -92,23 +92,22 @@ public final class Tracer {
     }
 
     /**
-     * In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId
-     * stack, so we just make one up (just in time). This matches the behaviour of Tracer#startSpan.
+     * In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId stack, so we just
+     * make one up (just in time). This matches the behaviour of Tracer#startSpan.
      * <p>
-     * n.b. this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
+     * n.b. this is a bit funky because calling maybeGetTraceMetadata multiple times will return different spanIds
      */
-    public static TraceMetadata getTraceMetadata() {
+    public static Optional<TraceMetadata> maybeGetTraceMetadata() {
         Trace trace = checkNotNull(currentTrace.get(), "Unable to getTraceMetadata when there is trace in progress");
 
         if (trace.isObservable()) {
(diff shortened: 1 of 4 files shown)


markelliot:
👍 from me, but I think one of the maintainers should probably also say yes

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -91,8 +91,14 @@ private static boolean shouldObserve(Observability observability) {
         throw new SafeIllegalArgumentException("Unknown observability", SafeArg.of("observability", observability));
     }
 
+    /**
+     * In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId
+     * stack, so we just make one up (just in time). This matches the behaviour of Tracer#startSpan.
+     * <p>
+     * n.b. this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
+     */
     @Beta
```
iamdanfox: I think if we're making this public can we drop the @Beta pls?

iamdanfox: Also will need to make the TraceMetadata interface itself public?

robert3005: fixed it up

carterkozak:
I'm curious what you're tracing where something like [OkhttpTraceInterceptor](https://github.com/palantir/tracing-java/blob/2150bd2886d36adb9ba078de350d1d31f50f8ba9/tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java#L50-L54) is insufficient.

> It's hard/impossible to integrate tracing with non okhttp http libraries as there's no way to access the same metadata as okhttptraceinterceptor has access to.

It can be done by adding an additional span using Tracer.startSpan nested within attaching a detached span -- unfortunately this creates an additional traced span, but everything else falls into place without expanding the public API.

markelliot:
We'd like to add tracing to dialogue, which supports many potential client targets, not just OkHttp, and has facility for managing this within dialogue instead of within the client targets.

As we've discussed many times, Tracer.startSpan doesn't return something that is easily closed on an asynchronous thread (Dialogue is futures-based), and DetachedSpan doesn't yield a spanId, parentSpanId, or originatingSpanId.

The proposal in #276 would handle this use-case pretty trivially, but it's not something that seems likely to be adopted some time soon.

A heavier alternative than this change is:
- split Dialogue's Channel into a dialogue-api project
- add a tracing-dialogue project that implements the channel
- make dialogue-core depend on tracing-dialogue

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -91,8 +91,13 @@ private static boolean shouldObserve(Observability observability) {
...
     }
 
-    @Beta
-    static TraceMetadata getTraceMetadata() {
+    /**
+     * In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId
+     * stack, so we just make one up (just in time). This matches the behaviour of Tracer#startSpan.
+     * <p>
+     * n.b. this is a bit funky because calling getTraceMetadata multiple times will return different spanIds
+     */
```
carterkozak, line 97: Let's document that callers must have prior knowledge that there's an active trace with a root span, otherwise this throws.

Should we return Optional<TraceMetadata> instead of throwing to avoid misuse?

carterkozak:
Thanks for the context, that sounds reasonable.

PR #400, opened 2020-01-14
Provide Tracing support for the full request lifecycle

carterkozak:
Implement TracedRequestHandler which can be applied prior to blocking
operations in order to capture more of the request lifecycle.
The existing handler attaches the detached span created by the
request handler in order to provide more specific timing information
describing the blocking portion of the request.

For backwards compatibility, the existing TracedOperationHandler
can still be used without the TracedRequestHandler, however the
detached request span will be started when the TracedOperationHandler
is first executed.

## Before this PR
We did not support tracing the full request lifecycle across synchronous and asynchronous operations.

## After this PR
==COMMIT_MSG==
Provide Tracing support for the full Undertow request lifecycle
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/InternalTracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/InternalTracers.java
@@ -1,0 +1,34 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+/**
+ * Internal utilities meant for consumption only inside of the tracing codebase.
+ *
+ * Note: while this is internal, backwards compatibility should be preserved
+ * for a couple releases to prevent breaking libraries which resolve mis-matched
+ * tracing dependency versions.
+ */
+public final class InternalTracers {
+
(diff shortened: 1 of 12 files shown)


tracing/src/main/java/com/palantir/tracing/InternalTracers.java
```diff
@@ -0,0 +1,34 @@
...
+package com.palantir.tracing;
+
+/**
+ * Internal utilities meant for consumption only inside of the tracing codebase.
+ *
+ * Note: while this is internal, backwards compatibility should be preserved
+ * for a couple releases to prevent breaking libraries which resolve mis-matched
+ * tracing dependency versions.
+ */
+public final class InternalTracers {
```
carterkozak, line 26: Not a huge fan of this, but I don't want these details to leak into the span objects because sampling shouldn't impact consumers.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -210,9 +210,27 @@ static DetachedSpan detachInternal(String operation, SpanType type) {
                 ? maybeCurrentTrace.getTraceId() : Tracers.randomId();
         boolean sampled = maybeCurrentTrace != null
                 ? maybeCurrentTrace.isObservable() : sampler.sample();
+        Optional<String> parentSpan = getParentSpanId(maybeCurrentTrace);
         return sampled
-                ? new SampledDetachedSpan(operation, type, traceId, getParentSpanId(maybeCurrentTrace))
-                : new UnsampledDetachedSpan(traceId);
+                ? new SampledDetachedSpan(operation, type, traceId, parentSpan)
+                : new UnsampledDetachedSpan(traceId, parentSpan);
```
carterkozak, line 216: Bugfix: parent span information was not propagated through unsampled detached spans.

carterkozak:
Pending discussion, needs more test coverage if we move forward with this.

carterkozak:
The other piece that would be helpful in this change is a component to bridge the exchange attachment into a servlet attribute that our jersey filter can take advantage of. Otherwise we cannot apply the full request tracing handler globally because we would get two separate server-incoming spans.

ferozco:
👍 Spoke offline, change looks good. We'll follow up in the undertow runtime to pick up the the handler

PR #561, opened 2020-06-24
RandomSampler provides a static factory, preferred over the constructor

carterkozak:
The factory takes advantage of deterministic values (zero and one)
to return singleton samplers which don't unnecessarily churn through
random data.

## After this PR
==COMMIT_MSG==
RandomSampler provides a static factory, preferred over the constructor to  advantage of deterministic values (zero and one).
==COMMIT_MSG==

## Possible downsides?
deprecation churn


iamdanfox (approved):

iamdanfox:
We can probably put a little refaster/error-prone thingy in WC to auto-fix the deprecated constructor -> static factory :)

PR #564, opened 2020-06-26
Fix: Move from javax to jakarta api jars

robert3005:
## Before this PR
Use javax jars that have been renamed to jakarta and all new development happens there

## After this PR
==COMMIT_MSG==
Switch javax dependencies to equivalent jakarta dependencies
==COMMIT_MSG==

## Possible downsides?
N/A


iamdanfox (approved):

carterkozak (approved):

versions.lock
```diff
@@ -87,6 +88,7 @@ io.dropwizard.metrics:metrics-jvm:4.0.2 (2 constraints: 241e5546)
 io.dropwizard.metrics:metrics-logback:4.0.2 (1 constraints: 7e0edc4f)
 io.dropwizard.metrics:metrics-servlets:4.0.2 (1 constraints: 400d371f)
 io.zipkin.java:zipkin:1.13.1 (1 constraints: 3805333b)
+javax.servlet:javax.servlet-api:3.1.0 (1 constraints: 830dcc28)
```
schlosna, line 91: I don't think this should still be here as it should be replaced by `jakarta.servlet:jakarta.servlet-api` above, right?

versions.lock
```diff
@@ -16,11 +16,12 @@ com.palantir.safe-logging:safe-logging:1.8.2 (3 constraints: 422aeb21)
...
 com.squareup.okio:okio:1.13.0 (1 constraints: 7f0cb509)
 io.undertow:undertow-core:2.0.17.Final (1 constraints: 54072861)
+jakarta.servlet:jakarta.servlet-api:4.0.4 (1 constraints: 0a050336)
+jakarta.ws.rs:jakarta.ws.rs-api:2.1.6 (1 constraints: 0b05fe35)
 javax.annotation:javax.annotation-api:1.2 (2 constraints: 2d21193d)
 javax.inject:javax.inject:1 (2 constraints: d614a0ab)
-javax.servlet:javax.servlet-api:3.1.0 (2 constraints: 88129b22)
 javax.validation:validation-api:1.1.0.Final (3 constraints: dc393f20)
-javax.ws.rs:javax.ws.rs-api:2.0.1 (8 constraints: ed9109e8)
+javax.ws.rs:javax.ws.rs-api:2.0.1 (7 constraints: e98c8a03)
```
schlosna, line 24: I don't think this should still be here as it should be replaced by `jakarta.ws.rs:jakarta.ws.rs-api` above, right?

schlosna (commented):
I think we either need to use the jakarta-renames plugin or define the few gradle module replacements used here

versions.lock
robert3005: this is coming from dropwizard via jetty

versions.lock
robert3005: you're right, let me look deeper into this

robert3005: https://github.com/palantir/tracing-java/pull/565 for a fix for this.

PR #574, opened 2020-07-13
Tracing uniquely identifies requests

carterkozak:
## Before this PR
Many requests may share a single traceId, there was no identifying information to distinguish between requests.

## After this PR
==COMMIT_MSG==
Tracing uniquely identifies requests
==COMMIT_MSG==

## Possible downsides?
* The request identifier is arguably not a tracing concern, implementing it at this layer is a matter of convenience.We already wrap requests and executors with tracing-aware components, and tell tracing when a request begins and ends, making tracing an ideal candidate to house this identifier.
* Possible performance cost of generating another unique identifier for each request.

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
@@ -18,6 +18,8 @@ package com.palantir.tracing.undertow;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tracing.DetachedSpan;
 import com.palantir.tracing.InternalTracers;
 import com.palantir.tracing.Observability;
@@ -73,6 +75,11 @@ final class UndertowTracing {
         // Populate response before proceeding since later operations might commit the response.
         exchange.getResponseHeaders().put(TRACE_ID, traceId);
         exchange.putAttachment(TracingAttachments.IS_SAMPLED, InternalTracers.isSampled(detachedSpan));
+        Optional<String> requestId = InternalTracers.getRequestId(detachedSpan);
+        if (!requestId.isPresent()) {
+            throw new SafeIllegalStateException("No requestId is set", SafeArg.of("span", detachedSpan));
+        }
+        exchange.putAttachment(TracingAttachments.REQUEST_ID, requestId.get());
         exchange.putAttachment(REQUEST_SPAN, detachedSpan);
         exchange.addExchangeCompleteListener(DetachedTraceCompletionListener.INSTANCE);
     }
--- a/tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracedRequestHandlerTest.java
+++ b/tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracedRequestHandlerTest.java
@@ -28,7 +28,7 @@ import com.palantir.tracing.api.SpanObserver;
 import com.palantir.tracing.api.TraceHttpHeaders;
 import io.undertow.Undertow;
 import io.undertow.server.HttpHandler;
-import io.undertow.server.handlers.ResponseCodeHandler;
+import io.undertow.util.HttpString;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URL;
@@ -70,7 +70,8 @@ public class TracedRequestHandlerTest {
 
         traceReportedLatch = new CountDownLatch(1);
         port = portSelector.incrementAndGet();
-        HttpHandler nextHandler = new TracedRequestHandler(ResponseCodeHandler.HANDLE_200);
+        HttpHandler nextHandler = new TracedRequestHandler(exchange -> exchange.getResponseHeaders()
+                .put(HttpString.tryFromString("requestId"), exchange.getAttachment(TracingAttachments.REQUEST_ID)));
         server = Undertow.builder()
                 .addHttpListener(port, null)
                 .setHandler(exchange -> {
@@ -101,6 +102,7 @@ public class TracedRequestHandlerTest {
         con.setRequestProperty(TraceHttpHeaders.TRACE_ID, "1234");
         assertThat(con.getResponseCode()).isEqualTo(200);
         assertThat(con.getHeaderField(TraceHttpHeaders.TRACE_ID)).isEqualTo("1234");
+        assertThat(con.getHeaderField("requestId")).isNotEmpty();
         assertThat(traceReportedLatch.await(5, TimeUnit.SECONDS)).isTrue();
         verifyNoMoreInteractions(traceSampler);
         verify(observer).consume(spanCaptor.capture());
--- a/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
+++ b/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
@@ -65,22 +65,6 @@ public interface DetachedSpan {
     }
 
     /**
-     * Marks the beginning of a span, which you can {@link #complete} on any other thread.
-     *
-     * @see DetachedSpan#start(String)
-     */
-    @CheckReturnValue
-    static DetachedSpan start(
-            Observability observability,
-            String traceId,
-            Optional<String> requestId,
-            Optional<String> parentSpanId,
-            String operation,
-            SpanType type) {
-        return Tracer.detachInternal(observability, traceId, requestId, parentSpanId, operation, type);
-    }
-
-    /**
      * Equivalent to {@link Tracer#startSpan(String, SpanType)}, but using this {@link DetachedSpan} as the parent
      * instead of thread state.
      */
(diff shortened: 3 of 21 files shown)


carterkozak:
needs more test coverage

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -64,6 +64,22 @@ static DetachedSpan start(
...
+    /**
+     * Marks the beginning of a span, which you can {@link #complete} on any other thread.
+     *
+     * @see DetachedSpan#start(String)
+     */
+    @CheckReturnValue
+    static DetachedSpan start(
+            Observability observability,
+            String traceId,
+            Optional<String> requestId,
```
carterkozak: This method should be unused, the system can generate a new requestId if this is a SERVER_INCOMING span.

carterkozak: delete

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -230,6 +278,31 @@ static DetachedSpan detachInternal(
...
+        }
+        if (detachedSpan instanceof UnsampledDetachedSpan) {
+            return ((UnsampledDetachedSpan) detachedSpan).requestId;
+        }
+        throw new SafeIllegalStateException("Unknown span type", SafeArg.of("detachedSpan", detachedSpan));
+    }
+
+    /** Returns the unique request identifier for this thread's trace. */
+    public static Optional<String> getRequestId() {
+        return checkNotNull(currentTrace.get(), "There is no trace").getRequestId();
```
carterkozak: I don't want requestId to be part of the public API here, perhaps this can be hidden.

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -37,6 +38,9 @@
      */
     public static final String TRACE_SAMPLED_KEY = "_sampled";
 
+    /** The key under which tracing request ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
+    public static final String REQUEST_ID_KEY = "_requestId";
```
carterkozak, line 42: This will appear in every log line. Do we want to abbreviate it in some way? As written it's self-explanatory.

ferozco: I like it as is

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -138,12 +141,42 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
                 .map(observable -> observable ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE)
```
ferozco, line 141: Can't leave a comment above, but we should update the javadoc so that we don't recommend people use a deprecated endpoint

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -138,12 +141,42 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
...
     }
 
-    /** Initializes the current thread's trace, erasing any previously accrued open spans. */
+    /**
+     * Initializes the current thread's trace, erasing any previously accrued open spans.
+     *
+     * @deprecated Use {@link #initTraceWithSpan(Observability, String, String, SpanType)}
+     */
+    @Deprecated
     public static void initTrace(Observability observability, String traceId) {
```
ferozco, line 153: There are a few usages of this method left in our tests, can we get rid of them as well?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -138,12 +141,42 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
...
 
-        setTrace(createTrace(observability, traceId));
+        setTrace(createTrace(observability, traceId, Optional.empty()));
     }
 
-    /** Initializes the current thread's trace, erasing any previously accrued open spans. */
+    /**
+     * Initializes the current thread's trace, erasing any previously accrued open spans.
+     *
+     * @deprecated Use {@link #initTraceWithSpan(Observability, String, String, SpanType)}
```
ferozco, line 150: there seems to be relatively little usage of this method, with the majority of it being in tests

tracing/src/test/java/com/palantir/tracing/TraceTest.java
```diff
@@ -21,19 +21,20 @@
...
+import java.util.Optional;
 import org.junit.Test;
 
 public final class TraceTest {
     private static final String ORIGINATING_SPAN_ID = "originating span id";
 
     @Test
     public void constructTrace_emptyTraceId() {
-        assertThatThrownBy(() -> Trace.of(false, "")).isInstanceOf(IllegalArgumentException.class);
+        assertThatThrownBy(() -> Trace.of(false, "", Optional.empty())).isInstanceOf(IllegalArgumentException.class);
```
ferozco, line 32: lots of churn haha

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: Ah you're right, they're left in tests that suppress deprecation :-/

tracing-undertow/src/test/java/com/palantir/tracing/undertow/TracedRequestHandlerTest.java
```diff
@@ -70,7 +70,8 @@ public void before() {
 
         traceReportedLatch = new CountDownLatch(1);
         port = portSelector.incrementAndGet();
-        HttpHandler nextHandler = new TracedRequestHandler(ResponseCodeHandler.HANDLE_200);
+        HttpHandler nextHandler = new TracedRequestHandler(exchange -> exchange.getResponseHeaders()
+                .put(HttpString.tryFromString("requestId"), exchange.getAttachment(TracingAttachments.REQUEST_ID)));
```
ferozco, line 74: should we make this `_requestId` for consistency and to avoid potential clobbering

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: Unfortunately many tests attempt to peel back layers of abstraction rather than testing results. These methods are only really used for server-incoming requests where we have prior knowledge of a request identifier.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: linked list style migration docs!

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
```diff
@@ -73,6 +75,11 @@ private static void setExchangeState(HttpServerExchange exchange, DetachedSpan d
         // Populate response before proceeding since later operations might commit the response.
         exchange.getResponseHeaders().put(TRACE_ID, traceId);
         exchange.putAttachment(TracingAttachments.IS_SAMPLED, InternalTracers.isSampled(detachedSpan));
+        Optional<String> requestId = InternalTracers.getRequestId(detachedSpan);
+        if (!requestId.isPresent()) {
+            throw new SafeIllegalStateException("No requestId is set", SafeArg.of("span", detachedSpan));
+        }
+        exchange.putAttachment(TracingAttachments.REQUEST_ID, requestId.get());
```
ferozco, line 82: I think we should add a response header as well for easier debugging

tracing/src/main/java/com/palantir/tracing/Tracer.java
ferozco: Sorry my comment was unclear, I was looking at internal usage across the fleet

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
carterkozak: I'm not sure I'm ready to modify the wire api at this point, the traceid is more actionable in the majority of cases as an api consumer while the requestId gives us the ability to do stronger structured analysis.

tracing/src/test/java/com/palantir/tracing/TraceTest.java
carterkozak: Agreed, I felt safer adding another param to all the test cases since it means we can't forget to pass the requestId along in production code.

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
ferozco: ok, we can add it later if needed

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: Ah, that makes sense :+1:

ferozco:
👍 looks great!

PR #624, opened 2020-10-06
Add `Tracer.hasUnobservableTrace()` API to simplify consumers

carterkozak:
This makes the code easier to follow and avoids additional unnecessary threadlocal map lookups.

==COMMIT_MSG==
Add `Tracer.hasUnobservableTrace()` API to simplify tracing framework implementations.
==COMMIT_MSG==

## Possible downsides?
none


ferozco:
👍

PR #678, opened 2021-01-20
log when clearing trace

tpetracca:
We're seeing instances of trace information disappearing from logs on a thread that previously had traceIds. Trying to track down where/how they're getting cleared.

==COMMIT_MSG==
debug logging when clearing trace
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -689,9 +689,17 @@ public final class Tracer {
 
     @VisibleForTesting
     static void clearCurrentTrace() {
+        logClearingTrace();
         currentTrace.remove();
         MDC.remove(Tracers.TRACE_ID_KEY);
         MDC.remove(Tracers.TRACE_SAMPLED_KEY);
         MDC.remove(Tracers.REQUEST_ID_KEY);
+    }
+
+    private static void logClearingTrace() {
+        log.debug("Clearing current trace", SafeArg.of("maybeTrace", Optional.ofNullable(currentTrace.get())));
+        if (log.isTraceEnabled()) {
+            log.trace("Stacktrace at time of clearing trace", new SafeRuntimeException("not a real exception"));
+        }
     }
 }
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -657,6 +657,8 @@ public final class Tracer {
         MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
         setTraceSampledMdcIfObservable(trace.isObservable());
         setTraceRequestId(trace.getRequestId());
+
+        logSettingTrace(trace);
     }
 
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -689,9 +689,17 @@ private static Trace getOrCreateCurrentTrace() {
...
     static void clearCurrentTrace() {
+        logClearingTrace();
         currentTrace.remove();
         MDC.remove(Tracers.TRACE_ID_KEY);
         MDC.remove(Tracers.TRACE_SAMPLED_KEY);
         MDC.remove(Tracers.REQUEST_ID_KEY);
     }
+
+    private static void logClearingTrace() {
+        log.debug("Clearing current trace", SafeArg.of("maybeTrace", Optional.ofNullable(currentTrace.get())));
```
carterkozak: nit: Let's wrap this with log.isDebugEnabled to prevent optional and safeArg allocations when debug logging is disabled.

note: no reason to use an Optional for the safe-arg, as far as the logging system is concerned empty optional and null are indistinguishable.

carterkozak: While we're here, thoughts on mirroring the debug line after a traceId is set? That way we get a full sense of the scope transitions.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -697,7 +705,9 @@ static void clearCurrentTrace() {
     }
 
     private static void logClearingTrace() {
-        log.debug("Clearing current trace", SafeArg.of("maybeTrace", Optional.ofNullable(currentTrace.get())));
+        if (log.isDebugEnabled()) {
+            log.debug("Clearing current trace", SafeArg.of("trace", currentTrace.get()));
+        }
         if (log.isTraceEnabled()) {
             log.trace("Stacktrace at time of clearing trace", new SafeRuntimeException("not a real exception"));
         }
```
carterkozak, line 711: we could move this conditional inside of the `isDebugEnabled` check -- if debug isn't enabled, trace cannot be enabled.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -678,6 +680,12 @@ private static void setTraceRequestId(Optional<String> requestId) {
...
     }
 
+    private static void logSettingTrace(Trace trace) {
+        if (log.isDebugEnabled()) {
+            log.debug("Setting trace", SafeArg.of("trace", trace));
```
carterkozak: nit: the trace object itself probably won't tell us a lot, I think we mostly care about traceId and isObservable, but those should exist in MDC.

If we remove the non-constant args, we no longer need an isDebugEnabled check
```suggestion
            log.debug("Setting trace");
```

carterkozak (approved):
oof, I never submitted comments. Sorry for the delay!

Approving, change looks good to me. Up to you whether you'd like to take action on the nits.

PR #681, opened 2021-01-29
add Tracers.wrapSupplier

rzpt:
## Before this PR

There is wrap(Callable) but that is suboptimal because Callable.call throws Exception.

## After this PR

==COMMIT_MSG==
Add Tracers.wrapSupplier to wrap a Supplier with tracing.
==COMMIT_MSG==

## Possible downsides?

None. The name is suboptimal because java can't decide if a lambda should be a callable or a supplier.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -164,6 +164,22 @@ public final class Tracers {
     }
 
     /**
+     * Wraps the given {@link Supplier} such that it uses the thread-local {@link Trace tracing state} at the time of
+     * it's construction during its {@link Supplier#get() execution}.
+     */
+    public static <V> Supplier<V> wrapSupplier(Supplier<V> delegate) {
+        return new TracingAwareSupplier<>(Optional.empty(), delegate);
+    }
+
+    /**
+     * Like {@link #wrap(Supplier)}, but using the given {@link String operation} is used to create a span for the
+     * execution.
+     */
+    public static <V> Supplier<V> wrapSupplier(String operation, Supplier<V> delegate) {
+        return new TracingAwareSupplier<>(Optional.of(operation), delegate);
+    }
+
+    /**
      * Wraps the given {@link Callable} such that it uses the thread-local {@link Trace tracing state} at the time of
      * it's construction during its {@link Callable#call() execution}.
      */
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -453,6 +469,34 @@ private static void restoreTrace(Optional<Trace> trace) {
...
+     * that constructs the {@link TracingAwareSupplier} instance rather than the thread that executes the supplier.
+     *
+     * <p>The constructor is typically called by a tracing-aware executor service on the same thread on which a user
+     * creates {@link Supplier delegate}, and the {@link #get()} method is executed on an arbitrary (likely different)
+     * thread with different {@link Trace tracing state}. In order to execute the task with the original (and
+     * intuitively expected) tracing state, we remember the original state and set it for the duration of the
+     * {@link #get() execution}.
+     */
+    @SuppressWarnings("deprecation")
+    private static class TracingAwareSupplier<V> implements Supplier<V> {
```
carterkozak, line 476: This is interesting, I would have expected this to add an operation around each time `Supplier.get()` is invoked, but not to capture the trace from the creation point because it's uncommon that suppliers are used to cross thread boundaries. I think suppliers are often called many times, where callables are (typically) called once.

e.g.
```java
private static class TracingSupplier<V> implements Supplier<V> {

  private final String operation;
  private final Supplier<V> delegate;
  
  TracingSupplier(String operation, Supplier<V> delegate) {
     this.operation = operation;
     this.delegate = delegate;
  }
  
  @Override
  public V get() {
    try (CloseableTracer ignored = CloseableTracer.startSpan(operation) {
      return delegate.get();
    }
  }
}
```

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -195,6 +195,22 @@ public static Runnable wrap(String operation, Runnable delegate) {
         return new TracingAwareRunnable(Optional.of(operation), delegate);
     }
 
+    /**
+     * Wraps the given {@link Supplier} such that it uses the thread-local {@link Trace tracing state} at the time of
+     * it's construction during its {@link Supplier#get() execution}.
+     */
+    public static <V> Supplier<V> wrapSupplier(Supplier<V> delegate) {
+        return new TracingAwareSupplier<>(Optional.empty(), delegate);
+    }
```
carterkozak: Most of the methods which don't take an operation name exist for historical reasons, ideally we'd require an operation name for new calls.

rzpt: Will remove.

tracing/src/main/java/com/palantir/tracing/Tracers.java
rzpt: Hm... In newer java style, any of the Functional types are likely be used multiple times, so maybe it would be worth it to change all of them?

carterkozak: I believe the runnable and callable wrappers are used by the executorservice wrappers where it's important to pass along the submitting threads traceId.

For functional references, I wouldn't necessarily want to force devs to drop type information in order to add tracing, e.g. replace a call from`FooFactory.getFoo()` with `Supplier<Foo>.get()`. I wonder if we should support something like a lighter weight tritium proxy to retain types?

Perhaps it would help if you can show me where you want to use this internally?

PR #691, opened 2021-03-01
Use SOURCE retention for immutables style annotations

pkoenig10:
See https://github.com/immutables/immutables/issues/291.

Example of similar fix in https://github.com/palantir/conjure-java-runtime-api/pull/369.

This eliminates compilation warnings for consumers that do not directly depend on immutables.


CRogers (approved):
:+1: Thanks for this!

PR #698, opened 2021-03-29
Decrease default sample rate from 1% to .05% to match internal tools

carterkozak:
## Before this PR
More sampled traces.

## After this PR
==COMMIT_MSG==
Decrease default sample rate from 1% to .05% to match internal tools
==COMMIT_MSG==

## Possible downsides?
Fewer sampled traces. That's the point though.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -67,7 +67,7 @@ public final class Tracer {
     private static volatile Consumer<Span> compositeObserver = _span -> {};
 
     // Thread-safe since stateless
-    private static volatile TraceSampler sampler = RandomSampler.create(0.01f);
+    private static volatile TraceSampler sampler = RandomSampler.create(0.005f);
 
     /** Creates a new trace, but does not set it as the current trace. */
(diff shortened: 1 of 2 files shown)


fawind (approved):

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -67,7 +67,7 @@ private Tracer() {}
     private static volatile Consumer<Span> compositeObserver = _span -> {};
 
     // Thread-safe since stateless
-    private static volatile TraceSampler sampler = RandomSampler.create(0.01f);
+    private static volatile TraceSampler sampler = RandomSampler.create(0.005f);
```
tpetracca: missing a zero

tpetracca (changes requested):

tracing/src/main/java/com/palantir/tracing/Tracer.java
```suggestion
    private static volatile TraceSampler sampler = RandomSampler.create(0.0005f);
```

carterkozak: ah damn, thanks!

tpetracca (approved):

fawind:
Oof, good catch. Thanks Tom!

PR #711, opened 2021-04-09
Wire tag metadata throughout the project

carterkozak:
This is the simpler of our available options, which leverages the
existing metadata map when spans are closed rather than replacing
the operation name string with a tuple of string and map.
Using many of our abstractions, this allows us to include *result*
data in tag values rather than only what we know up front. I've
added a simple status code tag for conjure-undertow, however
we could include all safe request headers, safe path portions,
and more. That should occur separately from the initial wiring
though.
However, many of our wrapper utility functions don't map so
nicely to this model, and take both operation name as well
as a map upon construction. It may be helpful to create a
type which encapsulates the two.

## TagRecorder details

The "low level" api for applying tags is optimized to prevent unnecessary,
and unintentional overhead both when tracing is enabled, and when it's
disabled.

Given a function with data `MyComplexObject` where we want to apply tags `foo` from `myComplexObject.getFoo()` and `bar` from `myComplexObject.getBar()`, we don't need to construct a map or even allocate a lambda:

```java
void func() {
  Tracer.fastStartSpan("loading a MyComplexObject");
  MyComplexObject object = loadMyComplexObject();
  Tracer.fastCompleteSpan(MyComplexObjectRecorder.INSTANCE, object);
}

private enum MyComplexObjectRecorder implements TagRecorder<MyComplexObject> {
    INSTANCE;

    <T> void record(TagAdapter<T> sink, T target, MyComplexObject data) {
        sink.tag(target, "foo", data.getFoo());
        sink.tag(target, "bar", data.getBar());
    }
}
```

A second `Map<String,String>` API is provided for ease of use outside of hot paths. This
is generally sufficient outside of inner-loop code:

```java
void func() {
  Tracer.fastStartSpan("loading a MyComplexObject");
  MyComplexObject object = loadMyComplexObject();
  Tracer.fastCompleteSpan(ImmutableMap.of("foo", object.getFoo(), "bar", object.getBar()));
}
```

In practice the above code delegates the call to the first API using an internal `TagRecorder<Map<String,String>>` implementation.

## Before this PR
no tags

## After this PR
==COMMIT_MSG==
Wire tag metadata throughout tracing-java
==COMMIT_MSG==

## Possible downsides?
more api surface. Not the easiest to use.


tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
```diff
@@ -101,14 +103,28 @@ public void exchangeEvent(HttpServerExchange exchange, NextListener nextListener
...
-                    detachedSpan.complete();
+                    detachedSpan.complete(statusMetadata(exchange.getStatusCode()));
                 }
             } finally {
                 nextListener.proceed();
             }
         }
     }
 
+    private static final ImmutableMap<String, String> METADATA_200 = ImmutableMap.of("status", "200");
```
jkozlowski: Is this some sort of standard field?

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -17,13 +17,14 @@
...
 import com.palantir.tracing.api.SpanType;
+import java.util.Map;
 
 /**
  * Wraps the {@link Tracer} methods in a closeable resource to enable the usage of the try-with-resources pattern.
  *
  * <p>Usage: try (CloseableTracer trace = CloseableTracer.start("traceName")) { [...] }
  */
-public final class CloseableTracer implements AutoCloseable {
+public class CloseableTracer implements AutoCloseable {
```
jkozlowski, line 27: Uuuu, I wonder why this is...

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -305,12 +304,18 @@ static boolean isSampled(DetachedSpan detachedSpan) {
     }
 
     private static final class SampledDetachedSpan implements DetachedSpan {
+        private static final int NOT_COMPLETE = 0;
+        private static final int COMPLETE = 1;
+
+        private static final AtomicIntegerFieldUpdater<SampledDetachedSpan> completedUpdater =
```
jkozlowski, line 318: Feels like a separate PR?

tracing/src/test/java/com/palantir/tracing/CloseableTracerTest.java
```diff
@@ -53,4 +58,26 @@ public void startsAndClosesSpanWithType() {
         }
         assertThat(Tracer.getAndClearTrace().top()).isEmpty();
     }
+
```
jkozlowski, line 61: Can you add a test for something that uses the lower level API? mostly for documentation?

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
carterkozak, line 27: Despite not being final, the private ctor ensures that this cannot be extended.

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
```diff
@@ -101,14 +102,34 @@ public void exchangeEvent(HttpServerExchange exchange, NextListener nextListener
...
+
+        static String statusString(int statusCode) {
+            // handle common cases quickly
+            switch (statusCode) {
+                case 200:
+                    return "200";
+                case 204:
+                    return "204";
+            }
+            return Integer.toString(statusCode);
```
jkozlowski, line 133: Hmm what's the impact of potentially allocating here?

jkozlowski: I feel like I remember the jvm interns up to some number automatically?

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -88,7 +93,7 @@ public DeferredTracer(Optional<String> operation) {
         this(operation.orElse(DEFAULT_OPERATION));
     }
 
-    public DeferredTracer(String operation) {
+    public DeferredTracer(String operation, Map<String, String> metadata) {
```
carterkozak, line 96: I opted against including a TagRecorder constructor for DeferredTracer because there's no guarantee that implementations will be serializable, which is important for spark jobs. Additionally, these are most frequently used for executors where the tags are constant across a single executor instance.

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -46,8 +54,38 @@ public static CloseableTracer startSpan(String operation, SpanType spanType) {
...
+    public static <T> CloseableTracer startSpan(
+            String operation, TagRecorder<? super T> recorder, T data, SpanType spanType) {
+        Tracer.fastStartSpan(operation, spanType);
+        if (!Tracer.isTraceObservable() || recorder.isEmpty(data)) {
+            return INSTANCE;
+        }
+        return new TaggedCloseableTracer<>(recorder, data);
+    }
+
     @Override
```
jkozlowski, line 82: Would it be useful to add a close that takes args? I feel like this one is maybe the most used one, so might be useful to be able to add args?

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
jkozlowski: Waaait how does that work: it's not private when it's in the same file? today I learned a new trick!

tracing/src/main/java/com/palantir/tracing/Tracer.java
jkozlowski: Did you run some benchmarks or something?

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -69,7 +70,33 @@ static DetachedSpan start(
...
+    /**
+     * Equivalent to {@link #childSpan(String, SpanType)}, but using {@link Map metadata} tags.
+     */
+    @MustBeClosed
+    default CloseableSpan childSpan(String operationName, Map<String, String> metadata, SpanType type) {
+        return childSpan(operationName, MapTagRecorder.INSTANCE, metadata, type);
+    }
+
+    @MustBeClosed
+    <T> CloseableSpan childSpan(String operationName, TagRecorder<? super T> recorder, T data, SpanType type);
```
carterkozak: Note I did not update the `completeAndStartChild` methods defined below (not visible in the default GH diff) because it's not obvious that the `operationName` argument would be associated with the child trace, while tags would be associated with its parent.

It's easy enough to implement that logic as needed, and produces more obvious code:
```java
 try (CloseableSpan child = detached.childSpan(operationName, type)) {
    detached.complete(<tags>);
    // work    
}
 ```

File: tracing/src/main/java/com/palantir/tracing/CloseableTracer.java (hunk 7 above)
Comment (jkozlowski) [reply tracing/src/main/java/com/palantir/tracing/CloseableTracer.java:82]: Hmm I guess you can also pass mutable data arg to to the constructor

File: tracing/src/main/java/com/palantir/tracing/TagRecorder.java [hunk 9]
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Interface which allows tracing spans to be provided with tag metadata. This interface
+ * is structured such that implementations can be created once and reused, given {@link S data}
+ * from which all tag values are extracted. Unlike other approaches, this doesn't require per-span
+ * lambdas and has minimal overhead in both the sampled and unsampled cases. This allows unsampled spans
+ * to result in zero allocation.
+ *
+ * @param <S> Stateful object type, an instance of which is passed through method calls with the {@link TagRecorder}.
+ */
+public interface TagRecorder<S> {
```
carterkozak: Naming is a little funny here:

Tracing has had an (unused) concept of `map<string,string>` metadata for years, so I retained the metadata name to describe the map, however here I call this a TagRecorder rather than a MetadataRecorder. I'd appreciate a second opinion. Maybe I rename all the `metadata` map parameter names to `tags`?

I bobbled between a few suffixes before using `Recorder`. I think recorder is probably fine, but `Translator` could also make sense. What do you think?

jkozlowski (commented):
First batch of comments/questions from reading in the browser; I'll give it a proper read as well

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
carterkozak: Nope, I just figured it would be helpful to tag requests with a status code since it allows us to quickly differentiate successes from failures and demonstrate how the tagging feature works. I've updated the implementation a tad since this comment.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: This was a total driveby (but I think the only one in this PR). An attempt to reclaim some of the heap bloat from additional fields in objects.

tracing-undertow/src/main/java/com/palantir/tracing/undertow/UndertowTracing.java
carterkozak: Overhead is minimal, but easy enough to avoid entirely in the most common cases, so I took that route.

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
carterkozak: Definitely wouldn't recommend passing mutable data, heh.
Unfortunately a close method with args wouldn't be compatible with how the CloseableTracer is most often used, within a try-with-resources block. I don't want folks to give up that safety

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
carterkozak: heh, it's a little wonky. Our FinalClass errorprone rule has to check for this case https://github.com/palantir/gradle-baseline/blob/0f716d5ec6063b4dae06b5f761f252178aac3941/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/FinalClass.java#L100-L102

tracing/src/test/java/com/palantir/tracing/CloseableTracerTest.java
carterkozak: I've added a test which uses the API, but it's not very good documentation since it defines the recorder inline where in practice a single recorder would be reused (although to be fair, the inner recorder is still far more efficient than the map codepath if objects aren't being reused)

carterkozak: I updated the description of this PR describing the new abstraction, hopefully that's more helpful.

carterkozak:
Dialogue example here: https://github.com/palantir/dialogue/pull/1208

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -120,4 +120,13 @@ private static Observability getObservabilityFromHeader(ContainerRequestContext
             return "1".equals(header) ? Observability.SAMPLE : Observability.DO_NOT_SAMPLE;
         }
     }
+
+    private enum ContainerResponseContextTagRecorder implements TagRecorder<ContainerResponseContext> {
+        INSTANCE;
+
+        @Override
+        public <T> void record(TagAdapter<T> sink, T target, ContainerResponseContext data) {
+            sink.tag(target, "status", Integer.toString(data.getStatus()));
```
jkozlowski: Seeing the dialogue one made me think: would it not make sense to have a #tag method that takes some more common values too? like int/long/boolean? or just object?

tracing/src/main/java/com/palantir/tracing/TagRecorder.java
jkozlowski: I think Recorder is good, maybe Translator vibes more with me, mostly because of this precedent: https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/EventTranslator.java#L27.

I would also maybe say TagAdapter -> TagSink

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -35,6 +36,13 @@ public static CloseableTracer startSpan(String operation) {
         return startSpan(operation, SpanType.LOCAL);
     }
 
+    /**
+     * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
+     */
+    public static CloseableTracer startSpan(String operation, Map<String, String> metadata) {
```
jkozlowski, line 42: I know there's precedent, but shouldn't this be startLocalSpan?

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
jkozlowski: Actually just to set the record straight, I think I knew about this, but it always felt meh. But here it feels rather justified.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
jkozlowski: This is an interesting, quite subtle constraint with the Serializable thing: should we encode that somehow here? checkstyle/errorprone, or at least a comment?

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -151,6 +166,25 @@ public void close() {
...
+        private final Optional<Trace> originalTrace;
+
+        TaggedCloseableTrace(Optional<Trace> originalTrace, Map<String, String> metadata) {
+            this.metadata = metadata;
+            this.originalTrace = originalTrace;
+        }
+
+        @Override
+        public void close() {
+            Tracer.fastCompleteSpan(metadata);
```
jkozlowski, line 180: Can we extract a small method: 

```
completeAndClear(@Nullable Map<String, String> metadata) {
  Tracer.fastCompleteSpan(metadata);
  if (Tracer.hasTraceId()) {
                Tracer.getAndClearTrace();
  }
}
```

and use that here and DefaultCloseableTrace#close()?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -453,29 +499,30 @@ static void fastDiscardSpan() {
...
+        fastCompleteSpan(NoTagRecorder.INSTANCE, NoTagRecorder.INSTANCE);
     }
 
     /**
      * Like {@link #fastCompleteSpan()}, but adds {@code metadata} to the current span being completed.
-     *
-     * @deprecated Use {@link #fastCompleteSpan()}
      */
-    @Deprecated
     public static void fastCompleteSpan(Map<String, String> metadata) {
```
jkozlowski, line 520: @Nullable right? you've designed this to be a no-op if the metadata is null?

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
jkozlowski, line 96: Hear me out: shouldn't this be ```Map<String, Object>``? If you're being perf-conscious, a boxed Integer is surely better than String representation or that?

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -69,7 +70,33 @@ static DetachedSpan start(
      * instead of thread state.
      */
     @MustBeClosed
-    CloseableSpan childSpan(String operationName, SpanType type);
+    default CloseableSpan childSpan(String operationName, SpanType type) {
```
jkozlowski, line 75: Call me old-fashioned, but these default methods scare me just a little bit. And the amount of them

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -525,17 +572,31 @@ private static void notifyObservers(Span span) {
...
                 .spanId(openSpan.getSpanId())
                 .type(openSpan.type())
                 .parentSpanId(openSpan.getParentSpanId())
                 .operation(openSpan.getOperation())
                 .startTimeMicroSeconds(openSpan.getStartTimeMicroSeconds())
-                .durationNanoSeconds(System.nanoTime() - openSpan.getStartClockNanoSeconds())
-                .putAllMetadata(metadata)
-                .build();
+                .durationNanoSeconds(System.nanoTime() - openSpan.getStartClockNanoSeconds());
+        tag.record(SpanBuilderTagAdapter.INSTANCE, builder, state);
```
jkozlowski: So this is quite hard to validate and probably maintain (if you make the callers do this), but *I feel* like the contract you have with TagRecorder is that you should call #isEmpty()?

```
if (!tag.isEmpty(state)) {
            tag.record(SpanBuilderTagAdapter.INSTANCE, builder, state);    
        }
```

jkozlowski (commented):
I think I'm happy to approve this, I've stared at it long enough now. I am not going to push for it, because I've done the review now, but my preference would be to actually split out into a bunch of PRs:

1. Perf things (like moving the completed field to be AtomicUpdater). Easy and self-contained, and you could probably push a benchmark as well!
2. Driveby's: like extracting little named-classes in preference of anonymous lambdas and whatnot.
3. Push new API for the tag stuff, impl does nothing.
4. Push impl, with the tests.

The whole change is quite subtle, and whilst I think I covered everything I wanted, it is still possible we both missed something.

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
carterkozak: Hmm, I could go either way on this. I've repeated the Integer.toString piece a few times, for the same status code data. The current structure makes it obvious that everything is encoded into a string for the tracing system, where object may lead to confusion about the representation (how is ArrayList of [1,2,3] represented? not obvious that toString values aren't helpful in tracing, and may be more expensive than the tracing operation itself).

We have the flexibility to add those overloads later on using default methods if we need to.

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
carterkozak: It depends. Most uses of this class are for `LOCAL` spans, so I think it's helpful to make it obvious that's the simplest choice without necessarily understanding what the span types mean. Realistically there's not much reason for CloseableTracer to support other span types at all -- server incoming requires additional data that CloseableTracer doesn't take (is_sampled, traceId, parentSpanId, etc). I suppose it could work with client outgoing spans now that we have the Tracer.maybeGetMetadata method, but that's almost exclusively used in code we own.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
carterkozak: Hmm, boxing an integer outside of the default -128 -> 127 (my memory on exact ranges here is a bit fuzzy, still need coffee) default `int`->`Integer` cache isn't substantially better than int -> String, but valhalla will change that one day! I can make a similar argument to the overloads on TagAdapter about subtlety of object->string conversions and unexpected tag value data. I think it's safe to design around the assumption that tag values will always be strings.

This tracer type is generally used with static operation name/tag inputs, for example from wrapped executors where the name is created on initialization and passed to each deferredtracer.

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
jkozlowski: Well, deep down in the rabbithole, this just adds it to a conjure type, which eventually gets shipped as jackson... Anyway, probably not worth it, but it looked like it could be a useful extension.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
carterkozak: The unit test validates that objects can be serialized, but it wouldn't fail if we happened to pass serializable implementations. We wouldn't want to serialize the entire raw input in that case either... serialization is way too subtle, and I've enjoyed not having to work with it for the last year ;-)

An error-prone rule for serializable data would be nifty, I've been burned by many serialization failures! Not sure how we'd set up such a check to avoid false positives though.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
carterkozak: I think this method can be simplified to avoid the `getAndClear` branch when an original trace is present:
```java
Tracer.fastCompleteSpan(metadata);
if (originalTrace.isPresent()) {
    Tracer.setTrace(originalTrace.get());
} else if (Tracer.hasTraceId()) {
    Tracer.getAndClearTrace();
}
```

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
carterkozak: Ya, it's a balancing act.

tracing/src/main/java/com/palantir/tracing/TagRecorder.java
carterkozak: I borrowed the translator idea from disruptor ;-) I'll rename to translator since I think it provides more meaning.

> I would also maybe say TagAdapter -> TagSink

Hmm, sink doesn't feel quite right because it isn't a destination, so much as a pass-through to `T target` which is the sink. Although that may not help readability....

I definitely need to add more javadoc in this interface.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: You're right, we might as well do the empty check here.
If it is empty, it shouldn't cost anything to call `tag.record`. The `isEmpty` method allows us to quickly validate whether or not we need to keep track of the instance in some cases.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: I don't want to support nulls in the API/contract, but I prefer for observability code to be tolerant to null inputs. I think adding the `recorder.isEmpty(data)` check in the Span.Builder adapter would allow this to accept nulls, otherwise the Span object will NPE.

carterkozak:
> Perf things (like moving the completed field to be AtomicUpdater). Easy and self-contained, and you could probably push a benchmark as well!

Ya, this would have been better as another PR. Unfortunately stacking in GH isn't as easy as I'd like, and I got lazy pushing through the AtomicIntegerFieldUpdater change.

> Driveby's: like extracting little named-classes in preference of anonymous lambdas and whatnot.

I think all of these were required to ferry around the new translator/data objects. I may have added something that I've forgotten about though!

> Push new API for the tag stuff, impl does nothing.
> Push impl, with the tests.

I wouldn't want to merge the API with a no-op impl because it allows folks to start migrating to it, but lose data in practice. I agree that it's best to review API and impl changes separately. It's unfortunate that this project does not separate API from implementation at all, which I'd like to resolve in a future tracing v2 library, but who knows when we'll get to that :-/

jkozlowski (approved):

PR #713, opened 2021-04-12
[Prototype] sympathetic tagging API

carterkozak:
This API is a tad more difficult to understand than a simple
map, however it allows us to do zero work (zero allocations) in
the unsampled case, and avoid unnecessary hash tables in all
cases.

## Before this PR
Maps everywhere, in all cases.

## After this PR

==COMMIT_MSG==
sympathetic tagging API
==COMMIT_MSG==

## Possible downsides?
Harder to read. Most spans come from infra though.

### Alternatives

We could structure this differently, the easiest to understand is replacing the existing `Map<String,String>` with `Supplier<Map<String, String>>` or improving that slightly by passing through values using `Consumer<BiConsumer<String, String>>` for `sink -> sink.apply("tagName", "tagValue")`. However in those cases an object allocation is required for each interaction.  Such interfaces may be trivially implemented using the API in this PR.
I tend to prefer keeping infra code as light-weight as possible -- many services are large and cannot reliably C2 as much as one might expect.
blab: Is this API worth the readability tradeoff? ¯\_(ツ)_/¯


jkozlowski:
+1; I think this can potentially have a bunch of nice benefits for AtlasDB/timelock/etc. and as long as we provide a nicer API for those that don't necessarily care, this should be fine. It think it would be nice to just prove this out (by taking say https://github.com/palantir/atlasdb and changing it's calls here: https://github.com/palantir/atlasdb/blob/develop/atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/CloseableTrace.java#L46.

carterkozak:
Cool, I'll push the API out to everything that's still using `Map<String, String>`, and add a simpler API option for folks who prefer a lambda. Either `Supplier<Map>` e.g. `() -> ImmutableMap.of("a", "b")` or an interface of the shape `Consumer<BiConsumer<String, String>>` e.g. `result -> result.apply("a", "b")` I tend to prefer the latter, since it avoids maps entirely (we can drop through to write directly into the final Span object) while being easy to understand and follow once it has proper interface types.

jkozlowski:
I would think just Map (no lambdas no Suppliers -> basically what you have in your first PR) is probably simplest for people (I'd also probably want a comment there steering people to think about the fact that this allocates)? Basically you either want fast, or what everyone can understand?

carterkozak:
Map may be easier to understand, and we can certainly use it, but I think most folks don't spend a lot of time thinking about allocations. In most cases it's not necessary to understand (app code that doesn't run on hot paths) however adding a supplier means we can avoid building the hash table in hot paths for traces that aren't sampled.

I think it `supplier<map<string,string>>` isn't much harder for folks to understand, it allows them to write naive code which still performs well without trying very hard. It's worse if you can pre-compute and reuse an immutable tag map, however that case has substantial overlap with the performance optimized API which might as well be used.

jkozlowski:
ok, I think if we just write example of what the supplier<map> would look like vs. performant API, and decide then. I'd think if you're doing a supplier<map> then doing the perf one is not that big of a stretch. So let's just try it out!

carterkozak:
Closing this PR -- I've merged the improvements into #711 so it's all reviewable in one place.

PR #717, opened 2021-04-18
Implement helpers for TagTranslator ease of use

carterkozak:
Translators can be combined similarly to java functional
consumers using an `andThen` method.

Translator.isEmpty uses `@SuppressWarnings("unused")` rather than
an underscore to make IDE interactions easier when overriding
the method, this way the `data` variable name can be used as is.

==COMMIT_MSG==
TagTranslator quality of life improvements
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
+++ b/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
@@ -50,8 +50,13 @@ public interface TagTranslator<S> {
      * required to implement. It is always safe to return {@code false} even if
      * {@link #translate(TagAdapter, Object, Object)} may not produce any tags.
      */
-    default boolean isEmpty(S _data) {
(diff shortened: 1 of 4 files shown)


tracing/src/main/java/com/palantir/tracing/TagTranslator.java
```diff
@@ -50,10 +50,15 @@
      * required to implement. It is always safe to return {@code false} even if
      * {@link #translate(TagAdapter, Object, Object)} may not produce any tags.
      */
-    default boolean isEmpty(S _data) {
+    default boolean isEmpty(@SuppressWarnings("unused") S data) {
         return false;
     }
 
+    /** Returns a new composed {@link TagTranslator}. */
+    default TagTranslator<S> andThen(TagTranslator<? super S> after) {
```
jkozlowski, line 58: I'm assuming this will all get inlined nicely

jkozlowski (approved):

PR #718, opened 2021-04-20
Tracer gracefully handles null tags

carterkozak:
==COMMIT_MSG==
Tracer gracefully handles null tags. Not a problem, but will be helpful once we start adding more standard http tags from nullable sources.
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
+++ b/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
@@ -59,14 +59,9 @@ public interface TagTranslator<S> {
         return new CompositeTagTranslator<>(this, after);
     }
 
-    /**
-     * Tag adapter object which insulates the implementation of the underlying data structure from callers.
-     */
+    /** Tag adapter object which insulates the implementation of the underlying data structure from callers. */
     interface TagAdapter<T> {
-        /**
-         * Implementations should ignore calls with {@code null} key or value, however implementations aren't
-         * expected to tread these as nullable.
-         */
+        /** No information will be recorded if either {@code key} or {@code value} are {@code null}. */
         void tag(T target, String key, String value);
 
         void tag(T target, Map<String, String> tags);
--- a/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
+++ b/tracing/src/main/java/com/palantir/tracing/TagTranslator.java
@@ -63,6 +63,10 @@ public interface TagTranslator<S> {
      * Tag adapter object which insulates the implementation of the underlying data structure from callers.
      */
     interface TagAdapter<T> {
+        /**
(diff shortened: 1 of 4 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -604,7 +604,9 @@ private static void notifyObservers(Span span) {
 
         @Override
         public void tag(Span.Builder target, String key, String value) {
-            target.putMetadata(key, value);
+            if (key != null && value != null) {
+                target.putMetadata(key, value);
+            }
         }
 
         @Override
```
jkozlowski, line 612: Nulls in maps anyone? :)

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -604,7 +604,9 @@ private static void notifyObservers(Span span) {
...
         public void tag(Span.Builder target, String key, String value) {
-            target.putMetadata(key, value);
+            if (key != null && value != null) {
```
jkozlowski, line 607: Can you also update javadoc on the TagAdapter to explain the null behavior?

jkozlowski (commented):
Small tweaks

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: Updated.

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: The map case is a little bit special, I don't think we need to handle nulls in that case. The null key/value piece above isn't strictly required, but I'd rather not allow tracing uses to cause 5xx failures in production.

jkozlowski: Well, the map case delegates to putAll in a conjure generate thing, which disallows nulls, so that'll throw.

tracing/src/main/java/com/palantir/tracing/TagTranslator.java
```diff
@@ -63,6 +63,10 @@ default boolean isEmpty(@SuppressWarnings("unused") S data) {
      * Tag adapter object which insulates the implementation of the underlying data structure from callers.
      */
     interface TagAdapter<T> {
+        /**
+         * Implementations should ignore calls with {@code null} key or value, however implementations aren't
```
jkozlowski: "tread" -> "treat"?

I'm a bit confused by the language here: I would suggest writing it from the perspective of the user. "If either key or value are null, no information will be recorded, instead of throwing an NPE"

jkozlowski (approved):

PR #719, opened 2021-04-20
Add an API describing common/standard tags based on DD APM

carterkozak:
## Before this PR
Custom tag names that were only understood by humans.

## After this PR
==COMMIT_MSG==
Add an API describing common/standard tags based on DD APM
==COMMIT_MSG==

## Possible downsides?
Directly coupled to dd APM, however the keys themselves may changed later on into another standard relatively painlessly.

### DIFF
--- a/tracing-api/src/main/java/com/palantir/tracing/api/TraceTags.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/TraceTags.java
@@ -1,0 +1,95 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing.api;
+
+/**
+ * A set of standard tracing tags meant to be reused where applicable. These are based on the
+ * <a href="https://docs.datadoghq.com/logs/processing/attributes_naming_convention/#default-standard-attribute-list">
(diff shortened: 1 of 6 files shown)


tracing-api/src/main/java/com/palantir/tracing/api/TraceTags.java
```diff
@@ -0,0 +1,95 @@
...
+
+    // 'http.url_details.host' may contain unsafe data and is currently excluded out of caution.
+    /** The HTTP host part of the URL. */
+    public static final String HTTP_URL_HOST = "http.url_details.host";
+    /** The HTTP port part of the URL. */
+    public static final String HTTP_URL_PORT = "http.url_details.port";
+    /**
+     * The HTTP path part of the URL.
+     * When setting this tag, the caller takes responsibility for redaction, ensuring the path is a known
+     * template which does not include user-provided path paramter data.
```
jkozlowski: "parameter"

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -128,7 +129,7 @@ private static Observability getObservabilityFromHeader(ContainerRequestContext
 
         @Override
         public <T> void translate(TagAdapter<T> adapter, T target, ContainerResponseContext data) {
-            adapter.tag(target, "status", Integer.toString(data.getStatus()));
+            adapter.tag(target, TraceTags.HTTP_STATUS_CODE, Integer.toString(data.getStatus()));
```
jkozlowski, line 132: Now that makes me think we should've made TraceTag an interface, so that TraceTags is discoverable...

jkozlowski (commented):
There's a small typo

tracing-api/src/main/java/com/palantir/tracing/api/TraceTags.java
carterkozak:
```suggestion
     * template which does not include user-provided path parameter data.
```

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
carterkozak: Perhaps, although I expect most folks won't use any of these tags because they have more specific information. I think these will primarily be helpful for us :-)

jkozlowski (approved):

PR #722, opened 2021-04-21
Redact the HTTP_URL_HOST constant for now

carterkozak:
Better to remove now than later.

## After this PR

==COMMIT_MSG==
Redact the HTTP_URL_HOST constant for now
==COMMIT_MSG==

## Possible downsides?
technically a break.


tracing-api/src/main/java/com/palantir/tracing/api/TraceTags.java
```diff
@@ -70,8 +70,8 @@
     // The data would likely be unsafe to share anyhow.
 
     // 'http.url_details.host' may contain unsafe data and is currently excluded out of caution.
```
carterkozak, line 72: I wrote the comment, but neglected to comment the value. I like listing the constants in this file to match the upstream documentation rather than removing them entirely, that way it's obvious why certain constants aren't listed.

tpetracca (approved):

PR #723, opened 2021-04-21
Jersey tracing uses standardized tags

carterkozak:
Note that the operation name is still dynamic based on verb and path.

==COMMIT_MSG==
Jersey tracing uses standardized tags
==COMMIT_MSG==

### DIFF
--- a/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
+++ b/tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
@@ -27,6 +27,8 @@ import com.palantir.tracing.api.TraceHttpHeaders;
 import com.palantir.tracing.api.TraceTags;
 import java.io.IOException;
 import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import javax.annotation.Priority;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
@@ -60,10 +62,7 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
     // Handles incoming request
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        String path = Optional.ofNullable(uriInfo)
-                .map(ExtendedUriInfo::getMatchedModelResource)
-                .map(Resource::getPath)
-                .orElse("(unknown)");
+        String path = getPathTemplate();
 
         String operation = "Jersey: " + requestContext.getMethod() + " " + path;
         // The following strings are all nullable
@@ -102,7 +101,15 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
         MultivaluedMap<String, Object> headers = responseContext.getHeaders();
         if (Tracer.hasTraceId()) {
             String traceId = Tracer.getTraceId();
-            Tracer.fastCompleteSpan(ContainerResponseContextTagTranslator.INSTANCE, responseContext);
+            Tracer.fastCompleteSpan(FunctionalTagTranslator.INSTANCE, sink -> {
+                sink.accept(TraceTags.HTTP_STATUS_CODE, Integer.toString(responseContext.getStatus()));
+                sink.accept(TraceTags.HTTP_URL_PATH_TEMPLATE, getPathTemplate());
+                sink.accept(TraceTags.HTTP_METHOD, requestContext.getMethod());
+                Object requestId = requestContext.getProperty(REQUEST_ID_PROPERTY_NAME);
+                if (requestId instanceof String) {
+                    sink.accept(TraceTags.HTTP_REQUEST_ID, (String) requestId);
+                }
+            });
             headers.putSingle(TraceHttpHeaders.TRACE_ID, traceId);
         } else {
             // When the filter is called twice (e.g. an exception is thrown in a streaming call),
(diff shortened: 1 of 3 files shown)


tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -124,12 +131,19 @@ private static Observability getObservabilityFromHeader(ContainerRequestContext
...
+                .orElse("(unknown)");
+    }
+
+    private enum FunctionalTagTranslator implements TagTranslator<Consumer<BiConsumer<String, String>>> {
         INSTANCE;
 
         @Override
-        public <T> void translate(TagAdapter<T> adapter, T target, ContainerResponseContext data) {
-            adapter.tag(target, TraceTags.HTTP_STATUS_CODE, Integer.toString(data.getStatus()));
+        public <T> void translate(TagAdapter<T> adapter, T target, Consumer<BiConsumer<String, String>> data) {
```
jkozlowski, line 145: Alright that took a second to grok. If you end up wanting a lambda a bit more, perhaps you want to ship that as an overload?

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -102,7 +101,15 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
         MultivaluedMap<String, Object> headers = responseContext.getHeaders();
         if (Tracer.hasTraceId()) {
             String traceId = Tracer.getTraceId();
-            Tracer.fastCompleteSpan(ContainerResponseContextTagTranslator.INSTANCE, responseContext);
+            Tracer.fastCompleteSpan(FunctionalTagTranslator.INSTANCE, sink -> {
+                sink.accept(TraceTags.HTTP_STATUS_CODE, Integer.toString(responseContext.getStatus()));
+                sink.accept(TraceTags.HTTP_URL_PATH_TEMPLATE, getPathTemplate());
+                sink.accept(TraceTags.HTTP_METHOD, requestContext.getMethod());
+                Object requestId = requestContext.getProperty(REQUEST_ID_PROPERTY_NAME);
```
jkozlowski, line 108: The undertow typed API attachments is rather nicer...

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
```diff
@@ -175,7 +175,10 @@ public void testTraceState_setsResponseStatus() {
         Span span = spanCaptor.getValue();
         assertThat(span.getOperation()).isEqualTo("Jersey: POST /trace");
         assertThat(span.getMetadata())
-                .containsEntry(TraceTags.HTTP_STATUS_CODE, Integer.toString(response.getStatus()));
+                .containsEntry(TraceTags.HTTP_STATUS_CODE, Integer.toString(response.getStatus()))
```
jkozlowski, line 178: nit: #containsEntry vs comparing the whole thing? I always prefer hasSameEntriesAs.

jkozlowski (approved):

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
carterkozak: agreed

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
carterkozak: Yep, this is a design I had considered, I wanted to try it out internally here to see how I like it -- Jersey already allocates a ton of data, so the lambda allocation is a drop in the bucket, I think most application code is similar so that such an api is reasonable.

If we add it as an overload, I'd end up writing new interfaces rather than the raw java functional interfaces, and allow chaining invocations to support expression lambdas e.g.
```java
sink -> sink.tag(TraceTags.HTTP_STATUS_CODE, Integer.toString(responseContext.getStatus()))
            .tag(TraceTags.HTTP_URL_PATH_TEMPLATE, getPathTemplate())
            .tag(TraceTags.HTTP_METHOD, requestContext.getMethod());
```

jkozlowski: That's cool

tracing-jersey/src/test/java/com/palantir/tracing/jersey/TraceEnrichingFilterTest.java
carterkozak: I thought containsEntry chains were easiest to read, but I don't have much preference for these things. Should we update https://github.com/palantir/assertj-automation to replace one with the other?

jkozlowski: I meant more around one checks that you have some entries, and another checks that you have some entries AND NOTHING ELSE. That way if someone changes some implementation in an additive way, they are forced to update the test.

carterkozak: Ah I see, it was my goal to avoid breaking the test upon additive updates, but we could do an exhaustive test instead.

PR #725, opened 2021-04-23
Undertow tracing supports configurable root request spans

carterkozak:
This will be leveraged to reintroduce endpoint information
into the span name to allow aggregation across individual
endpoints, rather than _all_ endpoints.

==COMMIT_MSG==
Undertow tracing supports configurable root request spans
==COMMIT_MSG==

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedRequestHandler.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedRequestHandler.java
@@ -33,12 +33,21 @@ import io.undertow.server.HttpServerExchange;
  */
 public final class TracedRequestHandler implements HttpHandler {
 
+    private static final String DEFAULT_OPERATION_NAME = "Undertow Request";
+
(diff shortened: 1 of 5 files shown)


tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedRequestHandler.java
```diff
@@ -33,26 +33,35 @@
  */
 public final class TracedRequestHandler implements HttpHandler {
 
+    private static final String DEFAULT_OPERATION_NAME = "Undertow Request";
```
jkozlowski, line 36: is this so you don't break the API?

jkozlowski (approved):

tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracedRequestHandler.java
carterkozak: Yep

PR #727, opened 2021-04-29
Add @Safe annotations to tracer APIs

carterkozak:
==COMMIT_MSG==
Add `@Safe` annotations to tracer APIs
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
@@ -16,6 +16,7 @@
 
 package com.palantir.tracing;
 
+import com.palantir.logsafe.Safe;
 import com.palantir.tracing.api.SpanType;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -32,14 +33,14 @@ private CloseableTracer() {}
     /**
      * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
      */
-    public static CloseableTracer startSpan(String operation) {
+    public static CloseableTracer startSpan(@Safe String operation) {
```
jkozlowski, line 36: Can you explain what does this actually do?

carterkozak: It's only there for documentation, has no impact at present. Eventually I'd like to build static analysis which validates the provided data is not unsafe, this lays some foundation.

jkozlowski (approved):

PR #746, opened 2021-07-06
Tracing no longer depends on afterburner

carterkozak:
## Before this PR
Afterburner was used by the AsyncSlf4jSpanObserver, however that
observer implementaiton is never used in production and has
serious performance issues at scale. In the future, we should split
the observer implementation into a separate module to decouple the
tracing library from jackson modules (with the excpetion of
annotations, perhaps).
## After this PR

==COMMIT_MSG==
Tracing no longer depends on afterburner
==COMMIT_MSG==

## Possible downsides?
None


tracing/src/main/java/com/palantir/tracing/AsyncSlf4jSpanObserver.java
```diff
@@ -45,7 +44,6 @@
 public final class AsyncSlf4jSpanObserver extends AsyncSpanObserver {
```
iamdanfox, line 44: interestingly I actually can't find any references to this class internally... a separate PR could even do a cheeky deprecation/delete??

iamdanfox (approved):

tracing/src/main/java/com/palantir/tracing/AsyncSlf4jSpanObserver.java
carterkozak: Yep, as noted on the pr description ;-)

carterkozak: I don think it’s helpful to provide an oss reference implementation of some kind though, so my preference is the external module

PR #751, opened 2021-07-27
tracing-java require jdk11 (up from jdk8)

carterkozak:
## After this PR
==COMMIT_MSG==
tracing-java require jdk11 (up from jdk8)
==COMMIT_MSG==


jkozlowski (approved):
Some build failures, but makes sense

PR #753, opened 2021-08-02
Delete deprecated AsyncTracer

iamdanfox:
## Before this PR

AsyncTracer was added 2 years ago in this PR https://github.com/palantir/tracing-java/pull/117, but superseded by the DetachedSpan API. An internal sourcegraph search shows no usages.

I'd quite like to prune the unnecessary bits of tracing-java's API as it makes it easier to imagine what it would look like for the internals to use something different (e.g. opentelemetry). _Ideally_ we would have included it in the most recent 5.0.0 release..

## After this PR
==COMMIT_MSG==
The deprecated `AsyncTracer` class is deleted.
==COMMIT_MSG==

## Possible downsides?
- While I can't find any usages of this class in an internal codesearch, it's possible someone is still running an old version of a _jar_ which is compiled and uses this method. In practise, the only usage I know of was c-j-r... and this also cut over to the new DetachedSpan in 2019 https://github.com/palantir/conjure-java-runtime/pull/1205


carterkozak (approved):

changelog/@unreleased/pr-753.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: break
+break:
```
iamdanfox, line 2: my only other thought about calling this a `break` and cutting another major rev is would we want to also delete any other deprecated things (like `OkhttpTraceInterceptor`) perhaps??

or:

```
@Deprecated
    public static OpenSpan of(String operation, String spanId, SpanType type, Optional<String> parentSpanId) 
```

also

`AsyncSlf4jSpanObserver`

carterkozak: Major revs are cheap, we can dev a few times. If there are other breaks you’d like to make in the same major rev as this one we can merge them around the same time.

PR #754, opened 2021-08-02
Delete AsyncSlf4jSpanObserver

iamdanfox:
## Before this PR

A sourcegraph search shows the only usages of `AsyncSlf4jSpanObserver` are in archived repos (so no longer writeable), or abandoned personal copies of http-remoting.

There is however one usage in tritium's tests: https://github.com/palantir/tritium/blob/14084bccbaf34133af38d5da2c52b136e0c0b9aa/tritium-tracing/src/test/java/com/palantir/tritium/tracing/TracingInvocationEventHandlerTest.java#L71, which I'm deleting here https://github.com/palantir/tritium/pull/1172

I'd like to delete the parts of tracing-java's API that we no longer use so that any attempt to try out different internals (e.g. OpenTelemetry) is a bit easier.

## After this PR
==COMMIT_MSG==
Delete `AsyncSlf4jSpanObserver` - internally we construct a SpanObserver which uses log4j2 to ensure that spans are serialized and logged without blocking threads that produce spans.
==COMMIT_MSG==


carterkozak:
Does this allow us to update the slf4j dependency in the `tracing` module from `compile` (api) to `implementation`?

tracing-undertow/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation project(':tracing-api')
     implementation 'com.google.guava:guava'
     implementation 'com.palantir.safe-logging:preconditions'
+    implementation "org.slf4j:slf4j-api"
```
carterkozak: This shouldn't be necessary after the safe-logger migration, just in `testImplementation`, right?

iamdanfox: Yep, after merging in develop this disappeared

versions.lock
```diff
@@ -1,9 +1,4 @@
 # Run ./gradlew --write-locks to regenerate this file
-com.fasterxml.jackson:jackson-bom:2.12.3 (13 constraints: d21fde18)
-com.fasterxml.jackson.core:jackson-annotations:2.12.3 (18 constraints: fa20d0ea)
-com.fasterxml.jackson.core:jackson-core:2.12.3 (18 constraints: f5617cda)
-com.fasterxml.jackson.core:jackson-databind:2.12.3 (26 constraints: 8cf4d2b4)
-com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.12.3 (4 constraints: 1e3f528b)
```
carterkozak, line 6: nice

carterkozak (approved):

PR #769, opened 2021-08-30
Support attaching DetachedSpan state to a thread without creating children

carterkozak:
==COMMIT_MSG==
Support attaching DetachedSpan state to a thread without creating children
==COMMIT_MSG==

## Potential Downsides

Previously we tried to couple thread state transitions/associations directly with named spans to more easily discover and remediate tracing state leak. Using `@MustBeClosed` should make it harder to create asymmetric start/complete tracing structures, and we're clearly pushing more spans than we'd like. I'd rather relax our opinions in this project than further reduce tracing rate while pushing unhelpful data.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
+++ b/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
@@ -139,6 +139,17 @@ public interface DetachedSpan {
     }
 
     /**
+     * Attaches the current {@link DetachedSpan} state to the current thread without creating additional spans.
+     * This is useful when a long-lived {@link DetachedSpan} measures many smaller operations (like async-io)
+     * in which we don't want to produce spans for each task, but do need tracing state associated for logging
+     * and potential child traces.
+     * @apiNote This must be executed within a try-with-resources block, and the parent detached span must still be
+     * completed separately.
+     */
+    @MustBeClosed
+    CloseableSpan attach();
+
+    /**
      * Completes this span. After complete is invoked, other methods are not expected to produce spans, but they must
      * not throw either in order to avoid confusing failures.
      */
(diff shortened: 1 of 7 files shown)


carterkozak:
@smoorpal This is the piece that should enable the capability you were looking for in dialogue -- this way we can decouple the concerns around retaining trace state from span creation.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -138,6 +138,17 @@ default DetachedSpan childDetachedSpan(@Safe String operation) {
...
+    /**
+     * Attaches the current {@link DetachedSpan} state to the current thread without creating additional spans.
+     * This is useful when a long-lived {@link DetachedSpan} measures many smaller operations (like async-io)
+     * in which we don't want to produce spans for each task, but do need tracing state associated for logging
+     * and potential child traces.
+     * @apiNote This must be executed within a try-with-resources block, and the parent detached span must still be
+     * completed separately.
+     */
+    @MustBeClosed
+    CloseableSpan attach();
```
carterkozak, line 150: The `CloseableSpan` type is a little bit odd here because it's not actually creating a span. I played with a new (though functionally identical) CloseableContext type to make the naming cleaner, but it also made the api larger, so I've pushed reusing the existing type.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -404,10 +400,8 @@ public void testWrapFuture_failure() {
             ListenableFuture<String> traced = Tracers.wrapListenableFuture(
                     operationName, () -> Futures.immediateFailedFuture(new SafeRuntimeException("result")));
             assertThat(traced).isDone();
-            assertThat(observed).hasSize(2);
-            // Inner operation must complete first to avoid confusion
-            assertThat(observed.get(0)).extracting(Span::getOperation).isEqualTo(operationName + " initial");
-            assertThat(observed.get(1)).extracting(Span::getOperation).isEqualTo(operationName);
+            assertThat(observed).hasSize(1);
+            assertThat(observed.get(0)).extracting(Span::getOperation).isEqualTo(operationName);
```
carterkozak, line 404: Everyone using the `Tracers.wrapListenableFuture` functionality will get cleaner tracing spans (no more "initial" span) when they upgrade.

smoorpal (approved):

tpetracca (approved):

PR #770, opened 2021-08-31
Support detaching from thread state without creating a new span

carterkozak:
## Before this PR
spans + deferredtracer.

## After this PR
Removed unhelpful spans: executors, runnables, and callables wrapped without an operation name no longer produce spans with the generic name `DeferredTracer(unnamed operation)`. It's possible that these provided value in some scenarios, however we strongly recommend providing a relevant name instead. These generic spans tend to be unhelpful debugging when things go wrong.

Some products trace thousands of items, sampled requests can result in a lot of allocation overhead deep-copying span stacks. DetachedSpan only peeks at one span.

==COMMIT_MSG==
Allow detachment from thread state without creating a new span.
==COMMIT_MSG==

## Possible downsides?
This is confusing. The original span may close/end before the detached element, however that's already the case using DeferredTracer. This change allows us to leverage the nicer, more efficient DetachedSpan API as a replacement for DeferredTracer.

I could not update DeferredTracer because it has been used to transfer an entire span stack to another thread, which I do not intend to support with this model.


tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -68,6 +68,17 @@ static DetachedSpan start(
...
     }
 
+    /**
+     * Creates a {@link DetachedSpan} instance based on the current tracing state without adding a new span.
+     * If there is no tracing state present a no-op instance is returned. This is equivalent to {@link #attach()}.
+     *
+     * @see DetachedSpan#attach()
+     */
+    @CheckReturnValue
+    static DetachedSpan detach() {
```
carterkozak: It feels potentially wrong to return `DetachedSpan` here because it provides (several) methods to `complete()` the span, however these cannot be completed because they don't actually produce spans.

Perhaps we extract a super-interface from `DetachedSpan` called `Detached` or `DetachedContext` which provides all the detached-span methods except for completion?

tracing/src/main/java/com/palantir/tracing/NopDetachedSpan.java
```diff
@@ -0,0 +1,104 @@
...
+
+import com.palantir.tracing.api.SpanType;
+import java.util.Map;
+
+enum NopDetachedSpan implements DetachedSpan {
+    INSTANCE;
+
+    @Override
+    public CloseableSpan childSpan(String _operationName, SpanType _type) {
+        return NopCloseableSpan.INSTANCE;
```
carterkozak: These methods should probably re-roll the dice in the current context rather than being no-ops.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -244,7 +244,7 @@ static DetachedSpan detachInternal(@Safe String operation, SpanType type) {
         Optional<String> requestId = getRequestId(maybeCurrentTrace, type);
         return sampled
                 ? new SampledDetachedSpan(operation, type, traceId, requestId, parentSpan)
-                : new UnsampledDetachedSpan(traceId, requestId, parentSpan);
+                : new UnsampledDetachedSpan(traceId, requestId, Optional.empty());
```
carterkozak, line 247: This is a simplification for readability: parentSpan is always empty here, but it wasn't obvious reading the code.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -413,40 +436,105 @@ private void warnIfCompleted(String feature) {
...
+            Trace maybeCurrentTrace = currentTrace.get();
+            Trace newTrace = Trace.of(true, traceId, requestId);
+            // Push the DetachedSpan OpenSpan to provide the correct parent information
+            // to child spans created within the context of this attach.
+            // It is VITAL that this span is never completed, it exists only for attribution.
+            newTrace.push(openSpan);
+            setTrace(newTrace);
+            // Do not complete the synthetic root span, it simply prevents nested spans from removing trace state, and
+            // allows
+            return maybeCurrentTrace == null ? REMOVE_TRACE : () -> Tracer.setTrace(maybeCurrentTrace);
```
carterkozak: These methods are duplicated with the sampled implementation. Once I'm happier with the API I'll revisit this.

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -507,34 +507,42 @@ private static void restoreTrace(Optional<Trace> trace) {
...
+            this.metadata = metadata;
         }
 
         @Override
         public V call() throws Exception {
-            try (DeferredTracer.CloseableTrace ignored = deferredTracer.withTrace()) {
+            try (CloseableSpan ignored = detachedSpan.childSpan(operation, metadata)) {
                 return delegate.call();
+            } finally {
+                detachedSpan.complete();
```
carterkozak: These `complete` calls are entirely unnecessary, however the reuse of `DetachedSpan` suggests we should use them in case we're given a 'real' detached span -- also for copy/pasted code.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -59,13 +59,13 @@ public void before() {
 
         Tracer.setSampler(AlwaysSampler.INSTANCE);
         // Initialize a new trace for each test
-        Tracer.setTrace(Trace.of(true, "defaultTraceId", Optional.empty()));
+        Tracer.initTraceWithSpan(Observability.SAMPLE, "defaultTraceId", "rootOperation", SpanType.LOCAL);
```
carterkozak, line 64: Fixes the test, a sampled trace with zero spans is an illegal state that doesn't exist in the wild.

tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -69,45 +71,16 @@ static DetachedSpan start(
...
-    @MustBeClosed
-    <T> CloseableSpan childSpan(@Safe String operationName, TagTranslator<? super T> translator, T data, SpanType type);
-
-    /**
-     * Equivalent to {@link Tracer#startSpan(String)}, but using this {@link DetachedSpan} as the parent instead of
-     * thread state.
+     * Creates a {@link Detached} instance based on the current tracing state without adding a new span.
+     * Note that
+     *
+     * If there is no tracing state present a no-op instance is returned. This is the inverse of {@link #attach()}.
```
tpetracca:
```suggestion
     * Note that if there is no tracing state present a no-op instance is returned. This is the inverse of
     * {@link #attach()}.
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -314,6 +337,37 @@ static boolean isSampled(DetachedSpan detachedSpan) {
...
+            this.data = data;
+        }
+
+        @Override
+        public void close() {
+            Tracer.fastCompleteSpan(translator, data);
+            Trace originalTrace = original;
+            if (originalTrace != null) {
+                Tracer.setTrace(originalTrace);
+            }
```
tpetracca, line 367: why do we need the `originalTrace` assignment here?

tpetracca: oh I see this was just copied from further down in the diff here

carterkozak: When we apply a trace somewhere else in a deferred way, it's important that it doesn't destroy any existing tracing that's already happening with a potentially (but not necessarily) different traceId.

Basically the tracing state before this is applied should exactly match the state after it's closed.

tpetracca (approved):
I'm a fan of removing those wrapped spans with the unhelpful/generic names. Agreed there are aspects of this that are confusing, but that's already the case anyway and this feels strictly better.

PR #771, opened 2021-09-01
Remove unused package-private fastDiscardSpan

carterkozak:
==COMMIT_MSG==
Remove unused package-private `fastDiscardSpan`
==COMMIT_MSG==

iirc this is an artifact of the deleted AsyncTracer.


schlosna (approved):
Confirmed no usages from anywhere I can see

PR #778, opened 2021-09-15
Remove unused headers: `X-B3-ParentSpanId` and `X-OrigSpanId`

carterkozak:
`X-B3-ParentSpanId` isn't standard zipkin, and doesn't provide
any data we don't already have. The parent span of an incoming
request span is the value of the `X-B3-SpanId` header. There's
no reason for additional header values.

`X-OrigSpanId` is a custom header added to provide some amount of
tracing functionality when traces aren't sampled, however it
hasn't quite worked and isn't supported by alternative tracing
engines.

==COMMIT_MSG==
Remove unused headers: `X-B3-ParentSpanId` and `X-OrigSpanId`
==COMMIT_MSG==

### DIFF
--- a/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
@@ -59,10 +59,14 @@ public abstract class OpenSpan {
     /**
      * Returns the identifier of the 'originating' span if one exists.
      *
-     * @see TraceHttpHeaders
+     * @see TraceHttpHeaders#ORIGINATING_SPAN_ID
+     * @deprecated No longer used.
      */
-    @Value.Parameter
-    public abstract Optional<String> getOriginatingSpanId();
+    @Deprecated
+    // Intentionally does not set @Value.Default because the value is always empty.
+    public Optional<String> getOriginatingSpanId() {
+        return Optional.empty();
+    }
 
     /** Returns a globally unique identifier representing a single span within the call trace. */
     @Value.Parameter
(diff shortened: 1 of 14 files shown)


carterkozak:
@CRogers This PR should reduce some of the cruft discussed on https://github.com/palantir/witchcraft-java-logging/pull/46 :-)

tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
```diff
@@ -83,26 +87,25 @@ public static Builder builder() {
         return new Builder().startTimeMicroSeconds(getNowInMicroSeconds()).startClockNanoSeconds(System.nanoTime());
     }
 
-    /** Use this factory method to avoid allocate {@link Builder} in hot path. */
+    /**
+     * Use this factory method to avoid {@link Builder} allocations in hot paths.
+     *
+     * @deprecated Use the variant without an originating span id
+     */
+    @SuppressWarnings("InlineMeSuggester")
```
jkozlowski, line 95: What's this magic?

jkozlowski (approved):

tracing-api/src/main/java/com/palantir/tracing/api/OpenSpan.java
carterkozak: Error prone has an inlineme annotation to document replacements on undesirable call sites. We don’t want to automate this migration because there’s potentially other data that can be removed.

sfackler:
X-B3-ParentSpanId is absolutely a part of standard zipkin but it's only used in an older style of propagation, where the server doesn't make a new span but rather "joins" onto the span the client created.

https://github.com/openzipkin/b3-propagation#parentspanid

PR #782, opened 2021-09-27
Draft For-User-Agent implementation

OStevan:
## Before this PR

In this draft tracing-java propagates an additional header across requests which will contain the user agent which was originally used to send a request.

Questions:
1. Should users of the library be able to set it when opening a span (essentially setting this info through the tracing lib)?
2. Should conjure clients set this header when sending a request?

_Tests are missing_


tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
```diff
@@ -37,4 +37,6 @@
      */
     @Deprecated
     String ORIGINATING_SPAN_ID = "X-OrigSpanId";
+
+    String FOR_USER_AGENT = "X-For-User-Agent";
```
carterkozak, line 41: nit: `X-` prefixes were deprecated several years ago in [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648)

tracing/src/main/java/com/palantir/tracing/Tracers.java
```diff
@@ -48,6 +48,11 @@
      */
     public static final String REQUEST_ID_KEY = "_requestId";
 
+    /**
+     * The Key under which origin user agents are inserted into SLF4J {@link org.slf4j.MDC MDCs}.
+     */
+    public static final String FOR_USER_AGENT_KEY = "__forUserAgent";
```
carterkozak, line 54: I don't think this belongs in the MDC, we can find it based on the requestId

carterkozak (commented):
Took a quick scan over this (not terribly thorough). Before implementing this I think we should factor out common tracing state (information which applies to a trace, not individual spans) into a single object that's shared between Trace and DetachedSpan (e.g. traceId and requestId). That way features like this one don't require so much plumbing.

OStevan:
Closing in favor of #790

PR #787, opened 2021-10-14
Common trace state for tracing

OStevan:
## Before this PR

Based on comments from #782 I extracted a `CommonTraceState` which contains the requestId and traceId. I'm not sure about the following changes:
1. Should the `getTraceId` and `getRequestId` be removed completely?
2. Making the `CommonTraceState` serializable introduced the nullable field for reqeustId and I'm not sure if the getter which wraps this nullable can be changed with something smarter.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
+++ b/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
@@ -22,22 +22,27 @@ import static com.palantir.logsafe.Preconditions.checkNotNull;
 import com.google.common.base.Strings;
 import java.io.Serializable;
 import java.util.Optional;
+import javax.annotation.Nullable;
 
 /**
  * Class representing the state which is created for each {@link Trace}. Contains the globally non-unique identifier of
  * a trace and a request identifier used to identify different requests sent from the same trace.
  */
 final class CommonTraceState implements Serializable {
+    private static final long serialVersionUID = 1L;
+
     private final String traceId;
-    private final Optional<String> requestId;
+
+    @Nullable
+    private final String requestId;
 
     static CommonTraceState create(String traceId, Optional<String> requestId) {
         checkArgument(!Strings.isNullOrEmpty(traceId), "traceId must be non-empty");
         checkNotNull(requestId, "requestId should be not-null");
-        return new CommonTraceState(traceId, requestId);
+        return new CommonTraceState(traceId, requestId.orElse(null));
     }
 
-    private CommonTraceState(String traceId, Optional<String> requestId) {
+    private CommonTraceState(String traceId, String requestId) {
         this.traceId = traceId;
         this.requestId = requestId;
     }
@@ -57,7 +62,7 @@ final class CommonTraceState implements Serializable {
      * distinguish between requests with the same traceId.
      */
     Optional<String> getRequestId() {
-        return requestId;
+        return Optional.ofNullable(requestId);
     }
 
     CommonTraceState deepCopy() {
@@ -66,6 +71,6 @@ final class CommonTraceState implements Serializable {
 
     @Override
     public String toString() {
-        return "CommonTraceState{" + "traceId='" + traceId + '\'' + ", requestId=" + requestId.orElse(null) + '}';
+        return "CommonTraceState{" + "traceId='" + traceId + '\'' + ", requestId=" + requestId + '}';
     }
 }
--- a/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
+++ b/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
@@ -61,7 +61,8 @@ final class CommonTraceState implements Serializable {
      * each time a new trace is created with a SERVER_INCOMING root span. This is a convenience in order to
      * distinguish between requests with the same traceId.
      */
-    Optional<String> getRequestId() {
+    @Nullable
+    String getRequestId() {
         return Optional.ofNullable(requestId);
     }
 
--- a/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
+++ b/tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
@@ -22,11 +22,15 @@ import static com.palantir.logsafe.Preconditions.checkNotNull;
 import com.google.common.base.Strings;
 import java.util.Optional;
 
-public final class CommonTraceState {
+/**
+ * Class representing the state which is created for each {@link Trace}. Contains the globally non-unique identifier of
+ * a trace and a request identifier used to identify different requests sent from the same trace.
+ */
+final class CommonTraceState {
     private final String traceId;
     private final Optional<String> requestId;
 
-    public static CommonTraceState create(String traceId, Optional<String> requestId) {
+    static CommonTraceState create(String traceId, Optional<String> requestId) {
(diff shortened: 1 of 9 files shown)


tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -58,7 +58,7 @@
     private static final String DEFAULT_OPERATION = "DeferredTracer(unnamed operation)";
 
     @Nullable
-    private final String traceId;
+    private final CommonTraceState commonTraceState;
```
carterkozak: Let's increment `serialVersionUID` to `2L`. We don't need serialization backcompat here since it's primarily used in spark where we're guaranteed that libraries are always in sync, but the increment will help us fail loudly if cross-version ser/de is attempted.

tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.base.Strings;
+import java.io.Serializable;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Class representing the state which is created for each {@link Trace}. Contains the globally non-unique identifier of
+ * a trace and a request identifier used to identify different requests sent from the same trace.
+ */
+final class CommonTraceState implements Serializable {
```
carterkozak: Could probably simplify the name (and reference variable/field names) to `TraceState`/`traceState`.

tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
```diff
@@ -0,0 +1,76 @@
...
+        return Optional.ofNullable(requestId);
+    }
+
+    CommonTraceState deepCopy() {
+        return new CommonTraceState(traceId, requestId);
+    }
+
+    @Override
+    public String toString() {
+        return "CommonTraceState{" + "traceId='" + traceId + '\'' + ", requestId=" + requestId + '}';
```
carterkozak: pedantic nit:
```suggestion
        return "CommonTraceState{traceId='" + traceId + "', requestId=" + requestId + '}';
```

tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
```diff
@@ -0,0 +1,76 @@
...
+     * each time a new trace is created with a SERVER_INCOMING root span. This is a convenience in order to
+     * distinguish between requests with the same traceId.
+     */
+    Optional<String> getRequestId() {
+        return Optional.ofNullable(requestId);
+    }
+
+    CommonTraceState deepCopy() {
+        return new CommonTraceState(traceId, requestId);
+    }
```
carterkozak: This method is unnecessary because this type is immutable, we can pass along the same instance without copying it.

tracing/src/main/java/com/palantir/tracing/CommonTraceState.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * The request identifier of this trace.
+     *
+     * The request identifier is an implementation detail of this tracing library. A new identifier is generated
+     * each time a new trace is created with a SERVER_INCOMING root span. This is a convenience in order to
+     * distinguish between requests with the same traceId.
+     */
+    Optional<String> getRequestId() {
```
carterkozak: Since this is not public API, I wouldn't mind returning a nullable string instead. Tracing is often on the hottest paths, so avoiding the optional indirection may be helpful. We use the NullAway errorprone plugin to ensure internally that nulls are respected.
```suggestion
    @Nullable
    String getRequestId() {
```

If it were me, since this is strictly a data structure (doesn't do any logic) I'd remove the `get` prefixes from accessor methods, I think `requestId = state.requestId()` is easiest to read.

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -186,12 +203,13 @@ boolean isObservable() {
...
         Trace deepCopy() {
-            return new Sampled(new ArrayDeque<>(stack), getTraceId(), getRequestId());
+            return new Sampled(new ArrayDeque<>(stack), getCommonTraceState().deepCopy());
         }
 
         @Override
         public String toString() {
-            return "Trace{stack=" + stack + ", isObservable=true, traceId='" + getTraceId() + "'}";
+            return "Trace{stack=" + stack + ", isObservable=true, traceId='"
+                    + getCommonTraceState().getTraceId() + "'}";
```
carterkozak: nit: Might as well just use the state objects string representation directly: `"state=" + getCommonTraceState()`

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -117,15 +117,15 @@ static TraceMetadata getTraceMetadata() {
             return trace.top().map(openSpan -> TraceMetadata.builder()
                     .spanId(openSpan.getSpanId())
                     .parentSpanId(openSpan.getParentSpanId())
-                    .traceId(trace.getTraceId())
-                    .requestId(trace.getRequestId())
+                    .traceId(trace.getCommonTraceState().getTraceId())
+                    .requestId(trace.getCommonTraceState().getRequestId())
```
carterkozak: I'd probably keep around the `getTraceId()` and `getRequetId()` methods on `Trace` for conveience/readability. Not sure it buys us anything to deprecate the method. JIT gives us a budget for bytecode that can be inlined and optimized, so smaller methods (reusing common accessors) are more likely to be optimized in hot paths.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -273,8 +272,8 @@ static DetachedSpan detachInternal(
         // The current trace has no impact on this function, a new trace is spawned and existing thread state
         // is not modified.
         return shouldObserve(observability)
-                ? new SampledDetachedSpan(operation, type, traceId, requestId, parentSpanId)
-                : new UnsampledDetachedSpan(traceId, requestId, parentSpanId);
+                ? new SampledDetachedSpan(operation, type, CommonTraceState.of(traceId, requestId), parentSpanId)
+                : new UnsampledDetachedSpan(CommonTraceState.of(traceId, requestId), parentSpanId);
```
carterkozak: nit: Might as well extract state creation above, and pass the reference to either the sampled or unsampled implementation.

carterkozak (commented):
Generally looks great, thanks for putting it together! I've noted a few nits, once those are cleaned up, this should be good to merge :-)

tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -44,14 +44,11 @@
...
+    private final TraceState traceState;
 
-    private final Optional<String> requestId;
-
-    private Trace(String traceId, Optional<String> requestId) {
-        checkArgument(!Strings.isNullOrEmpty(traceId), "traceId must be non-empty");
-        this.traceId = traceId;
-        this.requestId = checkNotNull(requestId, "requestId");
+    private Trace(TraceState traceState) {
+        checkNotNull(traceState, "Common trace state should not be null");
```
carterkozak:
```suggestion
        checkNotNull(traceState, "Trace state must not be null");
```

carterkozak (approved):
lgtm, thanks!

PR #790, opened 2021-10-21
Introduce ForUserAgent

OStevan:
## Before this PR
When looking at the trace users needed to take a look at the whole trace to find out which service/FE-app started the request chain. With this PR the tracing library will propagate this information if set through the `ForUserAgent` header.

## After this PR

==COMMIT_MSG==
User agent specified inside `ForUserAgent` http header will be propagated to requests made in the same trace.
==COMMIT_MSG==

## Possible downsides?

Possibly slight perf degradation as there is one more value to send/receive for every http request.

## Concerns
1. Should we have a mechanism to populate `ForUserAgent` based on standard user agent or should it only be propagated only if explicitly set? Something like [this](https://github.com/palantir/tracing-java/pull/782/files#diff-3aa4fa03293a5b01b58ee1e0c4907a25cc0ed77b82a7a38aa2e94ed971c793deR151-R158)
2. Should there be a mechanism in tracing to set these headers or should client libraries like dialogue provide capability to set this header?


tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
```diff
@@ -30,6 +30,8 @@
     String SPAN_ID = "X-B3-SpanId";
     String IS_SAMPLED = "X-B3-Sampled"; // Boolean (either “1” or “0”, can be absent)
 
+    String FOR_USER_AGENT = "ForUserAgent";
```
carterkozak: nit: Let's match the proxy-style `For-*` style, and standard `User-Agent`:
```suggestion
    String FOR_USER_AGENT = "For-User-Agent";
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -153,19 +157,23 @@ public static void initTrace(Optional<Boolean> isObservable, String traceId) {
...
      */
+    @Deprecated
     public static void initTraceWithSpan(
             Observability observability, String traceId, @Safe String operation, String parentSpanId, SpanType type) {
         setTrace(createTrace(
                 observability,
                 traceId,
-                type == SpanType.SERVER_INCOMING ? Optional.of(Tracers.randomId()) : Optional.empty()));
+                type == SpanType.SERVER_INCOMING ? Optional.of(Tracers.randomId()) : Optional.empty(),
+                Optional.empty()));
```
carterkozak: Can we simplify this to delegate to the new overload with `Optional.empty()` `forUserAgent`?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -178,7 +186,45 @@ public static void initTraceWithSpan(
...
+        fastStartSpan(operation, type);
```
carterkozak, line 193: Can  this delegate as well?

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -53,6 +53,8 @@
 
     public static final String REQUEST_ID_PROPERTY_NAME = "com.palantir.tracing.requestId";
 
+    public static final String FOR_USER_AGENT_PROPERTY_NAME = "com.palantir.tracing.forUserAgent";
```
carterkozak: I'm not sure this is necessary since we don't need to read it elsewhere, probably not helpful to apply it to the jersey context.

tracing-jersey/src/main/java/com/palantir/tracing/jersey/TraceEnrichingFilter.java
```diff
@@ -68,27 +70,42 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
         // The following strings are all nullable
         String traceId = requestContext.getHeaderString(TraceHttpHeaders.TRACE_ID);
         String spanId = requestContext.getHeaderString(TraceHttpHeaders.SPAN_ID);
+        Optional<String> forUserAgent =
+                Optional.ofNullable(requestContext.getHeaderString(TraceHttpHeaders.FOR_USER_AGENT));
```
carterkozak: I think reading the for-user-agent value is a bit trickier. If the incoming request has a `For-User-Agent` header, we use that, otherwise we use the `Fetch-User-Agent` (because conjure-typescript cannot override the browser user-agent), and if that doesn't exist, we use the `User-Agent` value.

carterkozak (commented):
This should also set for-user-agent in the undertow tracing handlers which are more common in our apps than jersey these days.

tracing-api/src/main/java/com/palantir/tracing/api/TracingHeadersEnrichingFunction.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing.api;
+
+public interface TracingHeadersEnrichingFunction<T> {
```
carterkozak:
```suggestion
/** Function used to adapt the tracing library to any arbitrary request type. */
public interface TracingHeadersEnrichingFunction<T> {
```

tracing-okhttp3/src/main/java/com/palantir/tracing/OkhttpTraceInterceptor2.java
```diff
@@ -47,11 +48,15 @@ public Response intercept(Chain chain) throws IOException {
...
                     .header(TraceHttpHeaders.TRACE_ID, Tracer.getTraceId())
                     .header(TraceHttpHeaders.SPAN_ID, metadata.getSpanId())
-                    .header(TraceHttpHeaders.IS_SAMPLED, Tracer.isTraceObservable() ? "1" : "0")
-                    .build());
+                    .header(TraceHttpHeaders.IS_SAMPLED, Tracer.isTraceObservable() ? "1" : "0");
+            Optional<String> forUserAgent = Tracer.getForUserAgent();
+            if (forUserAgent.isPresent()) {
+                requestBuilder.header(InternalTraceHttpHeaders.FOR_USER_AGENT, forUserAgent.get());
+            }
+            return chain.proceed(requestBuilder.build());
```
carterkozak, line 54: Should this use the new TracingHeaderEnrichingFunction?

tracing-okhttp3/src/main/java/com/palantir/tracing/okhttp3/OkhttpTraceInterceptor.java
```diff
@@ -53,6 +55,10 @@ public Response intercept(Chain chain) throws IOException {
                 .header(TraceHttpHeaders.TRACE_ID, Tracer.getTraceId())
                 .header(TraceHttpHeaders.SPAN_ID, span.getSpanId())
                 .header(TraceHttpHeaders.IS_SAMPLED, Tracer.isTraceObservable() ? "1" : "0");
+        Optional<String> forUserAgent = Tracer.getForUserAgent();
+        if (forUserAgent.isPresent()) {
+            tracedRequest.header(InternalTraceHttpHeaders.FOR_USER_AGENT, forUserAgent.get());
+        }
```
carterkozak: here as well

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -750,6 +800,28 @@ public static String getTraceId() {
         return checkNotNull(currentTrace.get(), "There is no trace").getTraceId();
     }
 
+    /**
+     * Returns the forUserAgent propagated inside the trace.
+     */
+    public static Optional<String> getForUserAgent() {
+        Trace trace = currentTrace.get();
+        return trace == null ? Optional.empty() : trace.getForUserAgent();
+    }
```
carterkozak: Let's hold off making this data accessible -- initially I'd like to let it flow through calls and verify the values are what we expect in our request logs.

tracing/src/main/java/com/palantir/tracing/InternalTraceHttpHeaders.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.palantir.tracing;
+
+/**
+ * Internal header names meant for consumption only inside of the tracing codebase.
+ */
+public interface InternalTraceHttpHeaders {
+    String FETCH_USER_AGENT = "Fetch-User-Agent";
+}
```
carterkozak: This shouldn't be public if we can avoid it

tracing-api/src/main/java/com/palantir/tracing/api/TracingHeadersEnrichingFunction.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing.api;
+
+/** Function used to adapt the tracing library to any arbitrary request type. */
+public interface TracingHeadersEnrichingFunction<T> {
```
carterkozak: This class should live in the `tracing` module (implementation), not tracing-api.

carterkozak (approved):
lgtm!

carterkozak:
:popcorn::robot:

carterkozak:
Ah, we'll need another +1 here since the autorelease RC has my name on it...

jkozlowski:
👍

jkozlowski (approved):

PR #849, opened 2022-02-17
Trace threadlocal entry is never removed, only nulled

carterkozak:
This prevents threadlocal map entry churn when traces are added
and cleared, however it comes at the potential cost of larger
threadlocal maps when tracing is not used. We generally expect
everything to be traced, so this shouldn't impact us.

==COMMIT_MSG==
Trace threadlocal entry is never removed, only nulled
==COMMIT_MSG==

### BEFORE
```
Benchmark                             (observability)  Mode  Cnt    Score    Error  Units
TracingBenchmark.traceWithSingleSpan           SAMPLE  avgt    3  551.519 ± 57.893  ns/op
TracingBenchmark.traceWithSingleSpan    DO_NOT_SAMPLE  avgt    3  317.826 ± 14.467  ns/op
TracingBenchmark.traceWithSingleSpan        UNDECIDED  avgt    3  342.436 ± 38.777  ns/op
```

```
Benchmark                             (observability)  Mode  Cnt    Score     Error  Units
TracingBenchmark.traceWithSingleSpan           SAMPLE  avgt    3  540.388 ±  69.014  ns/op
TracingBenchmark.traceWithSingleSpan    DO_NOT_SAMPLE  avgt    3  359.699 ± 208.046  ns/op
TracingBenchmark.traceWithSingleSpan        UNDECIDED  avgt    3  362.240 ±  96.292  ns/op
```

### AFTER
```
Benchmark                             (observability)  Mode  Cnt    Score    Error  Units
TracingBenchmark.traceWithSingleSpan           SAMPLE  avgt    3  503.538 ± 20.695  ns/op
TracingBenchmark.traceWithSingleSpan    DO_NOT_SAMPLE  avgt    3  301.762 ± 16.082  ns/op
TracingBenchmark.traceWithSingleSpan        UNDECIDED  avgt    3  323.679 ± 82.884  ns/op
```

```
Benchmark                             (observability)  Mode  Cnt    Score    Error  Units
TracingBenchmark.traceWithSingleSpan           SAMPLE  avgt    3  513.711 ± 21.862  ns/op
TracingBenchmark.traceWithSingleSpan    DO_NOT_SAMPLE  avgt    3  297.156 ± 24.606  ns/op
TracingBenchmark.traceWithSingleSpan        UNDECIDED  avgt    3  319.645 ± 33.003  ns/op
```

### 20 threads BEFORE
```
Benchmark                             (observability)  Mode  Cnt     Score     Error  Units
[...]

PR #850, opened 2022-02-17
Tracer#clearCurrentTrace sets thread local current trace to null

schlosna:
## Before this PR
Profiles of a service showed many `java.lang.ThreadLocal$ThreadLocalMap$Entry` allocations in the `Tracer$UnsampledDetachedSpan.childSpan` read path, as threads using `Tracers.wrapWithNewTrace` had previously removed the thread local entry.
```
    com.palantir.tracing.Tracer$UnsampledDetachedSpan.childSpan(String, TagTranslator, Object, SpanType)
    java.lang.ThreadLocal.get()
    java.lang.ThreadLocal.setInitialValue()
    java.lang.ThreadLocal$ThreadLocalMap.set(ThreadLocal, Object)
```

We can adjust `Tracer#clearCurrentTrace` to set the thread local current trace to null so that the `ThreadLocal$ThreadLocalMap$Entry` is reused in these cases.

## After this PR

==COMMIT_MSG==
Tracer#clearCurrentTrace sets thread local current trace to null

Instead of removing the thread local current trace, we now set it to null to avoid the potentially expensive allocation of ThreadLocal$ThreadLocalMap$Entry when a thread is reused.
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -919,7 +919,7 @@ public final class Tracer {
     @VisibleForTesting
     static void clearCurrentTrace() {
         logClearingTrace();
-        currentTrace.remove();
+        currentTrace.set(null);
         MDC.remove(Tracers.TRACE_ID_KEY);
         MDC.remove(Tracers.TRACE_SAMPLED_KEY);
         MDC.remove(Tracers.REQUEST_ID_KEY);


carterkozak:
#849

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -919,7 +919,7 @@ private static Trace getOrCreateCurrentTrace() {
     @VisibleForTesting
     static void clearCurrentTrace() {
         logClearingTrace();
-        currentTrace.remove();
+        currentTrace.set(null);
         MDC.remove(Tracers.TRACE_ID_KEY);
         MDC.remove(Tracers.TRACE_SAMPLED_KEY);
         MDC.remove(Tracers.REQUEST_ID_KEY);
```
schlosna, line 925: I didn't touch these as I didn't see these pop in profiles, and log4j2 uses `org.apache.logging.log4j.spi.GarbageFreeSortedArrayThreadContextMap` & `org.apache.logging.log4j.util.SortedArrayStringMap` where the `MDC.remove` & `MDC.put` don't directly free/allocate

carterkozak: https://github.com/palantir/tracing-java/pull/849#issuecomment-1043269441

schlosna:
Closing in favor of #849

PR #959, opened 2022-08-08
Migrate to Jakarta Equivalents

mglazer:
## Before this PR

There was no jakarta equivalent for the libraries, this would prevent tracing-java from being used in modern services which upgrade to the Jakarta equivalents.

## After this PR

==COMMIT_MSG==
Introduced jakarta equivalents of all modules which had a javax dependency

Additionally had to do a few things:

1. Migrated from spotify's ADT to derive4j which generates types without a legacy javax.annotation.Generated type
2. Fixed a few tests
3. Upgraded newer libraries to Junit5
4. Legacy tests were deleted because keeping the versions pinned to the correct transitives was an exercise in futility.
==COMMIT_MSG==


settings.gradle
```diff
@@ -4,9 +4,15 @@ include 'tracing'
...
 include 'tracing-demos'
-include 'tracing-jaxrs'
-include 'tracing-jersey'
 include 'tracing-okhttp3'
-include 'tracing-servlet'
 include 'tracing-test-utils'
-include 'tracing-undertow'
+
+withJakarta('tracing-servlet')
+withJakarta('tracing-undertow')
```
carterkozak: Is there any jakarta involved with tracing-undertow?

tracing-undertow-jakarta/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'com.palantir.revapi'
+
+dependencies {
+    api project(':tracing')
+    api 'io.undertow:undertow-core:2.3.0.Alpha1'
```
carterkozak: I thought only undertow-servlet differed in 2.3.0 (jakarta by default) where undertow-core shouldn't be impacted.

build.gradle
```diff
@@ -41,6 +41,7 @@ apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.consistent-versions'
 apply plugin: 'com.palantir.baseline-java-versions'
 apply plugin: 'com.palantir.jdks.latest'
+apply plugin: 'com.palantir.consistent-versions'
```
carterkozak: repeated from line 41
```suggestion
```

mglazer: Sorry, still working through this.

tracing-servlet/build.gradle
```diff
@@ -17,16 +17,42 @@
...
+        version {
+            strictly '2.0.10'
+        }
+    }
     testImplementation "junit:junit"
     testImplementation 'org.assertj:assertj-core'
+
 }
+
+configurations.all {
```
mglazer: This is a little funky, but it's because dropwizard brings in a transitive dependency of Jersey, so we have to force it down this way (otherwise it's pulled up through GCV). Everything needs to be pulled up except for the inject package which has a different versioning scheme.

tracing-undertow-jakarta/build.gradle
mglazer: Oh maybe, let me look through it.

tracing-servlet/build.gradle
carterkozak: ouch, I should have replaced dropwizard with something like this: https://github.com/palantir/atlasdb/pull/5978
Not worth doing within this PR though

tracing-jaxrs/build.gradle
```diff
@@ -19,9 +19,13 @@ apply plugin: 'org.inferred.processors'
...
 apply plugin: 'com.palantir.revapi'
 
+versionsLock {
+    disableJavaPluginDefaults()
+}
+
 dependencies {
     api project(":tracing")
-    api "jakarta.ws.rs:jakarta.ws.rs-api"
+    api "javax.ws.rs:javax.ws.rs-api:2.1.1"
```
schlosna: can we put this in versions.props or do we explicitly have to pin the old javax deps?

tracing-jersey-jakarta/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'com.palantir.external-publish-jar'
+apply plugin: 'com.palantir.revapi'
+
+dependencies {
+    api "org.glassfish.jersey.core:jersey-server"
```
schlosna: what from `jersey-server` is needed for API? should this only need to expose JAX-RS API?

tracing-jersey/build.gradle
```diff
@@ -17,17 +17,40 @@
...
+        }
+    }
     implementation project(':tracing-api')
 
-    testImplementation "io.dropwizard:dropwizard-testing"
+    testImplementation "io.dropwizard:dropwizard-testing", {
+        version {
+            strictly '2.0.10'
+        }
+    }
```
schlosna: can we comment why these versions are strictly pinned?

tracing-undertow-jakarta/build.gradle
mglazer: Yup, you were right purged.

tracing-servlet/build.gradle
mglazer: I ended up having to just delete these tests...there were too many edge cases that ended up breaking because further in the dependency graph certain dependencies, like slf4j, hibernate, etc. were being pinned higher, and that was incompatible with these tests.

tracing-jaxrs/build.gradle
mglazer: This was partially stylistic just so it was explicit where the deps were coming from. Let me see if the latest build yells at me for anything (like unused pins).

tracing-jersey-jakarta/build.gradle
mglazer: my guess is:

```
import org.glassfish.jersey.server.ExtendedUriInfo;
import org.glassfish.jersey.server.model.Resource;
```

Does look like it's actually exposed on the API though, so I could move this to impl.

tracing-jersey/build.gradle
mglazer: yup

settings.gradle
mglazer: Not anymore.

tracing-jaxrs/build.gradle
```diff
@@ -19,9 +19,13 @@ apply plugin: 'org.inferred.processors'
...
 apply plugin: 'com.palantir.revapi'
 
+versionsLock {
+    disableJavaPluginDefaults()
+}
+
 dependencies {
     api project(":tracing")
-    api "jakarta.ws.rs:jakarta.ws.rs-api"
+    api "javax.ws.rs:javax.ws.rs-api"
```
schlosna, line 28: Should we `@Deprecate` everything in the non-Jakarta modules to flag these for migration?

tracing-jaxrs-jakarta/src/main/java/com/palantir/tracing/jaxrs/JaxRsTracers.java
```diff
@@ -0,0 +1,55 @@
...
+import com.palantir.tracing.Detached;
+import com.palantir.tracing.DetachedSpan;
+import com.palantir.tracing.Tracers;
+import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.core.StreamingOutput;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.concurrent.Callable;
+
+public final class JaxRsTracers {
```
schlosna, line 29: I'm assuming anything currently using `tracing-jaxrs` should migrate to `tracing-jaxrs-jakarta` and we'll deprecate/remove the non-Jakarta implementations in a future major breaking release, right?

schlosna (commented):
Overall LGTM, defer to @carterkozak on merge. 

I am assuming we'll want to minimize the window of dual javax/jakarta support, so wondering if we should open source the gradle-jakarta-renames plugin as a bridge until everything migrates artifact coordinates. Looks like there a few other stragglers with `javax` dependencies in non-test classpaths:

- https://github.com/palantir/sls-packaging/blob/0f4f12160c525ef7e4dc6ea8ec517801d2b0d89a/versions.lock#L23
- https://github.com/palantir/encrypted-config-value/blob/f477d33ac3d353528c3c874c44f9e22f58815c36/versions.lock#L50-L56
- https://github.com/palantir/gradle-baseline/blob/6452ab19fcdb7cfa9c85b8a85e6eeec0bc3abd4d/versions.lock#L47-L48
- https://github.com/palantir/cassandra-manager/blob/1b37306e09325521bec1f385c2054bfbca48add1/versions.lock#L51

tracing-jaxrs-jakarta/src/main/java/com/palantir/tracing/jaxrs/JaxRsTracers.java
mglazer: Sort of? Generally, people _should_ migrate as gradually more and more things will start breaking, to the point that eventually we can retire the legacy versions of these. But until that happens across _all_ libraries, we can't really say: You should do this.

Does that make sense?

tracing-jaxrs/build.gradle
mglazer: Maybe I follow up a commit to deprecate everything? I still have at least 6 other libraries that have to migrate first before I could even provide an alternative for someone to migrate to.

mglazer:
As for the question on open sourcing the internal plugin, I don't have strong opinions on this. However, given the seemingly endless set of edge cases, we ended up opting for the more discoverable: "Copy/Paste" approach which while a fair bit more work, ends up being easier to maintain in the long term for some of these stable libraries, ie: we can just delete directories wholesale in the future, rather than having to make further modifications to the legacy projects.

What we will very likely end up doing is writing automation to convert projects from the `tracing-java-jaxrs` to `tracing-java-jaxrs-jakarta`. That's conceivable, but again, isn't really attainable until we've migrated all upstream libraries first to produce compliant versions.

carterkozak (approved):
Great stuff, thank you!

mglazer:
Also, I realized I mis-spoke since we sort of have 2 jakarta-javax bridge plugins. The one you're thinking of probably makes sense to open source, but I think what I want to do first is add some functionality to it to ensure that component constraints are respected with regards to not having 2 tracing libraries on the classpath (for example).

PR #962, opened 2022-08-12
Add Jakarta EE9 Equivalents of all libraries

mglazer:
## After this PR

==COMMIT_MSG==
Add Jakarta EE9 Equivalents of all libraries

This is the second attempt at this. This version however is
significantly simpler than the previous version since we had
already converted all of the tests to a simpler undertow only
implementation, removing the Dropwizard dependency which
inadvertently bumped all of our dependencies past compatible
versions.

Tests have been consciously dropped in the non-jakarta versions as
the dependencies get messed up with those.

Similarly, revapi was dropped from the non-jakarta versions because of
a quirk in how that plugin works. It's not providing value here anyways though.

==COMMIT_MSG==


tracing-jaxrs-jakarta/src/main/java/com/palantir/tracing/jaxrs/JaxRsTracers.java
```diff
@@ -0,0 +1,55 @@
...
+import com.palantir.tracing.Detached;
+import com.palantir.tracing.DetachedSpan;
+import com.palantir.tracing.Tracers;
+import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.core.StreamingOutput;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.concurrent.Callable;
+
+public final class JaxRsTracers {
```
carterkozak, line 29: I wonder if we could get away with silently dropping this feature/module in the jakarta migration. I see three uses across two projects internally, which could be handled in a couple lines of code.

mglazer: Ya, and I think I know one of them. I'm indifferent, it's not causing pain though.

carterkozak: Ack, entirely up to you.

carterkozak (approved):

PR #986, opened 2022-11-04
Add enable flag to closeable tracers

raiju:
## Before this PR

Modifying `CloseableTracer`s to be conditionally enabled is impossible.

```
try (CloseableTracer trace = CloseableTracer.startSpan("foo")) {
   ...
}
```

would have to be written (e.g. when using logger setting to control trace levels):

```
boolean enableTracing = log.isDebugEnabled();

try {
   if (enableTracing) {
      Tracer.fastStartSpan("foo");
   }
   ...
} finally {
    if (enableTracing) {
        Tracer.fastCompleteSpan();
    }
}
```

which makes tracing mix with the actual code.

## After this PR

==COMMIT_MSG==
Add enable flag to closeable tracers
==COMMIT_MSG==

This enables the following code:

```
try (CloseableTracer trace = CloseableTracer.startSpan("foo", log.isDebugEnabled())) {
   ...
}
```

## Possible downsides?

Happy to discuss alternative arg ordering/method naming (should it be `startSpanIf` instead? Should the boolean be first?). The reasoning for the ordering was to keep the operation first for readability, but to keep the flag in a consistent place across overloads.

### DIFF
--- a/changelog/@unreleased/pr-986.v2.yml
+++ b/changelog/@unreleased/pr-986.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add enable flag to closeable tracers
+  links:
+  - https://github.com/palantir/tracing-java/pull/986
--- a/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
@@ -80,6 +80,73 @@ public class CloseableTracer implements AutoCloseable {
         return new TaggedCloseableTracer<>(translator, data);
     }
 
+    /**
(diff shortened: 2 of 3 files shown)


scapetta:
To be clear on how we expect this would be used:
We would have some spans that are generated anytime there is a trace (as is currently the case).
We would have other spans that would have to be turned on and we would do this by using an already existing config point in the service (if Logging = debug, set traceEnabled to true for example).

Correct?

carterkozak:
I’m not sure about this, let’s catch up when I’m not on airplane wifi :-)

raiju:
Taking an alternative approach (PR to follow!).

Short version: introduce a formal concept of trace levels & allow traces to be tagged as such.

PR #1007, opened 2023-03-01
trace locals

bavardage:
==COMMIT_MSG==
Introduce trace locals. Each trace gets its own independently settable copy of the variable. 
==COMMIT_MSG==


tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -39,6 +41,8 @@ final class TraceState implements Serializable {
     @Nullable
     private final String forUserAgent;
 
+    private final IdentityHashMap<TraceLocal<?>, Object> traceLocals;
```
carterkozak: We can use a ConcurrentHashMap and declare final equals+hashcode on `TraceLocal` which delegate to super (Object) for identity semantics.

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -64,6 +65,10 @@ private Tracer() {}
     // when our observers are modified.
     private static volatile Consumer<Span> compositeObserver = _span -> {};
 
+    private static final Map<String, TraceLocalObserver<?>> traceLocalObservers = new HashMap<>();
+    // see above comment on compositeObserver
+    private static volatile Consumer<TraceState> compositeTraceLocalObserver = _traceState -> {};
```
carterkozak: tracelocal listeners should be fully encapsulated within the TraceLocal object itself (An overridable method, or custom listener interface would handle this nicely)

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,38 @@
...
+    public TraceLocal(Supplier<T> initialValue) {
+        this.initialValue = initialValue;
+    }
+
+    @Nullable
+    public T get() {
+        return Tracer.getTraceLocalValue(this, initialValue);
+    }
+
+    public void set(T value) {
```
carterkozak: Possibly worth adding `remove()` to avoid `set(null)` calls.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.palantir.tracing;
+
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+public final class TraceLocal<T> {
+
+    private final Supplier<T> initialValue;
```
carterkozak: What are your expectations for the initialValue, how should it be used?
In a world with listeners, is there an expectation that listeners are invoked for a TraceLocal with an initial value which has never had `get` or `set` called within a given trace?

bavardage: yeah, maybe the potential for confusion isn't worth the convenience.

I was imagining: if get/set never called, then listeners are not invoked.

For convenience, it would be nice to have thread locals like `TraceLocal<Counter> counter = new TraceLocal(Counter::new)` which you could easily just use like `counter.get().inc()`...

bavardage: What do you think? Worth the potential for confusion?

tracing/src/main/java/com/palantir/tracing/Tracer.java
bavardage: yeah, fair enough - done

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: added a test to demonstrate current behaviour

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,102 @@
...
+            log.warn(
+                    "Overwriting existing TraceLocalObserver with name {} by new observer: {}",
+                    SafeArg.of("name", name),
+                    UnsafeArg.of("observer", observer));
+        }
+        if (observers.size() >= 5) {
+            log.warn("Five or more TraceLocalObservers registered: {}", SafeArg.of("observers", observers.keySet()));
+        }
+
+        observers.put(name, observer);
```
bavardage: i didn't do the computing the composite single function thing - thoughts on at what point that becomes worthwhile?

carterkozak: I think we can avoid it entirely by supporting exactly one listener interface when these are created. One could provide a listener which itself maintains a collection of observers that may be registered/unregistered over time for the behavior you've implemented, but this model would keep the minimal case less complex.

In terms of sequencing, I think I'd prefer if we split this feature into two PRs: One which adds the concept of a TraceLocal, and a second which expands them to support listeners.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
carterkozak: I think the initialValue component is reasonable, but should be documented to describe the edge cases

tracing/src/main/java/com/palantir/tracing/TraceLocalObserver.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+public interface TraceLocalObserver<T> {
+    void onTraceComplete(T value);
```
carterkozak: Should this include other attributes from the `RequestState` e.g. `traceId`, `optional<requestI>`, etc? Such values could also be included in the provided threadlocal value if necessary, but I suspect the traceId at minimum would be helpful for reporting (allows us to tie anything that we report back to the larger operation)

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: ok, i'll strip the listener stuff out

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: added javadoc (and will add more when adding observers)

tracing/src/main/java/com/palantir/tracing/TraceLocalObserver.java
bavardage: yeah makes sense, will add that in the second PR adding observers

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -374,6 +375,42 @@ static String getRequestId(DetachedSpan detachedSpan) {
...
+        Trace maybeCurrentTrace = currentTrace.get();
+        if (maybeCurrentTrace == null) {
+            return;
+        }
+
+        if (value != null) {
+            maybeCurrentTrace.getTraceState().getTraceLocals().put(traceLocal, value);
+        } else {
+            maybeCurrentTrace.getTraceState().getTraceLocals().remove(traceLocal);
+        }
```
schlosna: Is it worth pulling out traceLocals map to local since it is on both paths to shrink method bytecode size? Also nit on avoiding negation:

```suggestion
        Map<TraceLocal<?>, Object> traceLocals = maybeCurrentTrace.getTraceState().getTraceLocals();
        if (value == null) {
            traceLocals.remove(traceLocal);
        } else {
            traceLocals.put(traceLocal, value);
        }
```

tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -50,6 +54,7 @@ private TraceState(String traceId, @Nullable String requestId, @Nullable String
         this.traceId = traceId;
         this.requestId = requestId;
         this.forUserAgent = forUserAgent;
+        this.traceLocals = new ConcurrentHashMap<>();
```
schlosna: How much does this allocation add to perf overhead of tracing? Should we lazily initialize this map only when trace locals are used?

carterkozak: :+1: we should use an atomic-reference-field-updater to thread-safely initialize `traceLocals` to a new map as needed

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: Also `set` shouldn't `remove`, it should only set. Remove should be handled separately, there are subtle performance implications.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    @Nullable
+    public T get() {
+        return Tracer.getTraceLocalValue(this, initialValue);
+    }
+
+    /**
+     * Sets the value of this trace local for the current trace.
+     */
+    public void set(@Nonnull T value) {
```
carterkozak: Unlike ThreadLocal, this can race because traces are shared across threads. I wonder if we should provide some way to acknowledge replaced values by returning the previous value or taking some form of collector. Likely unnecessary in most cases, but it would be helpful to retain the capability.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Creates a trace local variable.
+     *
+     * When not already set, and this variable is accessed in a trace with the {@link #get()} method, the supplier
+     * will be invoked to supply a value (which will then be stored for future accesses).
+     *
+     * The supplier is thus normally invoked once per trace, but may be invoked again in case of subsequent
+     * invocations of {@link #remove()} followed by get.
+     */
+    public TraceLocal(Supplier<T> initialValue) {
```
carterkozak: Should we require a supplier in all cases? is `T` nullable?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -374,6 +375,42 @@ static String getRequestId(DetachedSpan detachedSpan) {
         throw new SafeIllegalStateException("Unknown span type", SafeArg.of("detachedSpan", detachedSpan));
     }
 
+    @Nullable
+    static <T> T getTraceLocalValue(TraceLocal<T> traceLocal) {
```
carterkozak: unused?

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -374,6 +375,42 @@ static String getRequestId(DetachedSpan detachedSpan) {
...
+    static <T> T getTraceLocalValue(TraceLocal<T> traceLocal, Supplier<T> initialValue) {
+        Trace maybeCurrentTrace = currentTrace.get();
+        if (maybeCurrentTrace == null) {
+            return null;
+        }
+
+        return (T) maybeCurrentTrace
+                .getTraceState()
+                .getTraceLocals()
+                .computeIfAbsent(traceLocal, _key -> initialValue.get());
```
carterkozak: Let's eagerly transform the `Supplier<T>` into `Function<?, T>` in the TraceLocal, that way we don't need to allocate lambdas for each `get` invocation.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
carterkozak: I think we should make the ctor private, and require a builder for TraceLocal initialization. That way we can optionally allow an initial-value-provider (so get operations don't add data to the internal map unnecessarily) and add the listener interface easily in a follow-up.

changelog/@unreleased/pr-1007.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: improvement
+improvement:
+  description: Introduce trace locals. Each trace gets its own independently settable
+    copy of the variable. Observers can be registered, which get called with the value
+    of the trace local when a trace completes.
```
carterkozak:
```suggestion
    copy of the variable.
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
bavardage: yeah makes sense

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: builder seems overkill for now, but I've made ctor private/added static factory methods

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: what do you mean by taking a collector?

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
carterkozak: How do you envision eventually creating a new TraceLocal with an initial value supplier and a listener?

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
carterkozak: either we return the previous value from `set`, or allow you to provide an update function which may take the previous value e.g. `(TraceLocal<Integer>) traceLocal.update(1, (previousValue, newValue) -> previous == null ? newValue : previousValue + newValue);` I think this sort of API is overkill though.

Edge cases returning the original value from `set` include:
* Do we return `initialValue.get()` if the value is unset?
* Should we provide an API to safely update a value if the current value matches our expectation (avoid update races)?
* Do we need a `setIfAbsent` for the common case, to avoid clobbering values across threads? Otherwise setting, and unsetting when our expectations arent met could allow mutations to be consumed by the temporary update value that we roll back.

tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -79,12 +84,17 @@ String forUserAgent() {
...
+
     @Override
     public String toString() {
         return "TraceState{"
                 + "traceId='" + traceId + "', "
                 + "requestId='" + requestId + "', "
-                + "forUserAgent='" + forUserAgent
-                + "'}";
+                + "forUserAgent='" + forUserAgent + "', "
+                + "traceLocals=" + traceLocals
```
carterkozak: We should not include the traceLocals map in toString because we have no guarantee that folks won't add tokens. Worth adding a test that traceLocal values aren't included in TraceState.toString
```suggestion
                + "traceLocals=" + traceLocals
```

tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -374,6 +375,43 @@ static String getRequestId(DetachedSpan detachedSpan) {
...
+            return;
+        }
+
+        Map<TraceLocal<?>, Object> traceLocals = maybeCurrentTrace.getTraceState().getTraceLocals();
+        if (value == null) {
+            traceLocals.remove(traceLocal);
+        } else {
+            traceLocals.put(traceLocal, value);
+        }
+    }
```
carterkozak: I think we can align the code with intent a bit better by exposing a package-private nullable getter for`TraceState` here, then implemeting the rest of this logic within TraceLocal.

tracing/src/main/java/com/palantir/tracing/TraceState.java
bavardage: done

tracing/src/main/java/com/palantir/tracing/Tracer.java
bavardage: it increases the potential for concurrency problems though I think - i.e. without being able to use computeIfAbsent on the ConcurrentHashMap

bavardage: could make tracelocal get synchronized?

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Creates a trace local variable.
+     *
+     * When not already set, and this variable is accessed in a trace with the {@link #get()} method, the supplier
+     * will be invoked to supply a value (which will then be stored for future accesses).
+     *
+     * The supplier is thus normally invoked once per trace, but may be invoked again in case of subsequent
+     * invocations of {@link #remove()} followed by get.
+     */
+    private TraceLocal(Supplier<T> initialValue) {
```
schlosna:
```suggestion
    private TraceLocal(@Nullable Supplier<T> initialValue) {
```

tracing/src/main/java/com/palantir/tracing/TraceState.java
bavardage: done

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: will return original value from set for now (documenting that this returns null if unset, not initialValue.get()). We can add update/setIfAbsent stuff in future if needed.

bavardage: (as in, those are easy additive changes)

tracing/src/main/java/com/palantir/tracing/Tracer.java
carterkozak: My preference would be to avoid splitting the complexity between classes, so `Tracer` provides an internal package-private `TraceState getTraceState()` and `TraceState` provides `@Nullable Map<> getTraceLocals()` and `Map<> getOrCreateTraceLocals()`. Then TraceLocal itself can do any threadsafe logic it likes on the map.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,149 @@
...
+        if (traceState == null) {
+            return null;
+        }
+
+        if (traceState.getTraceLocals() == null) {
+            // no trace locals ever set, short circuit (avoid creating the trace local map)
+            return null;
+        }
+
+        return (T) traceState.getOrCreateTraceLocals().remove(this);
```
carterkozak:
```suggestion
        Map<TraceLocal<?>, Object> traceLocals = traceState.getTraceLocals();
        if (traceLocals == null) {
            // no trace locals ever set, short circuit (avoid creating the trace local map)
            return null;
        }

        return (T) traceLocals.remove(this);
```

tracing/src/test/java/com/palantir/tracing/TraceTest.java
```diff
@@ -44,4 +44,14 @@ public void testToString() {
...
+
+    @Test
+    public void testToString_doesNotContainTraceLocals() {
+        Trace trace = Trace.of(true, TraceState.of("traceId", Optional.empty(), Optional.empty()));
+
+        TraceLocal<String> traceLocal = TraceLocal.of();
+        trace.getTraceState().getOrCreateTraceLocals().put(traceLocal, "secret-value");
+
+        assertThat(trace.toString()).doesNotContain("secret");
+    }
```
carterkozak, line 56: Nice

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,149 @@
...
+    @Nullable
+    private final Function<? super TraceLocal<?>, T> initialValue;
+
+    private TraceLocal(@Nullable Supplier<T> initialValue) {
+        if (initialValue == null) {
+            this.initialValue = null;
+        } else {
+            // eagerly transform supplier to avoid allocation per invocation
+            // (computeIfAbsent takes a Function)
+            this.initialValue = _ignored -> initialValue.get();
```
carterkozak: Let's add some nullness validation here -- always easier in an initial release than adding new validation later on.
```suggestion
            this.initialValue = _ignored -> Preconditions.checkNotNull(initialValue.get(), "TraceLocal initial value must not be null");
```

carterkozak (commented):
I'm happy with this pending the last couple of nits. @schlosna Anything from your end?

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: oh sure

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -0,0 +1,152 @@
...
+    private final Function<? super TraceLocal<?>, T> initialValue;
+
+    private TraceLocal(@Nullable Supplier<T> initialValue) {
+        if (initialValue == null) {
+            this.initialValue = null;
+        } else {
+            // eagerly transform supplier to avoid allocation per invocation
+            // (computeIfAbsent takes a Function)
+            this.initialValue = _ignored ->
+                    Preconditions.checkNotNull(initialValue.get(), "TraceLocal initial value must not be null");
```
schlosna, line 46: Can you confirm my understanding here -- if the supplier/function return null, we'll throw in `TraceLocal::get` not here in constructor? I think we could allow the supplier to return `null` and the value will just not be added to the `traceLocals` map ([`ConcurrentHashMap` allows this](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent(K,java.util.function.Function))).

```suggestion
            this.initialValue = _ignored -> initialValue.get();
```

schlosna (commented):
Overall seems good to me. Had one question in TraceLocal

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: yeah, already fixed in https://github.com/palantir/tracing-java/pull/1007/commits/ae573c1e4544f389654c20acbf75a3253edb3e52

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: yeah, throws in get() currently. Don't have strong feeling about what behaviour is best

carterkozak: My goal with this suggestion is to ensure folks aren't using a supplier which returns null instead of using the factory with no supplier, since that one is more efficient in some cases (`get()` on an empty trace doesn't initialize a new map). I think the behavior is more consistent for users this way as well, but I think we should add a checkNotNull in the `withInitialValue` static factory to throw on ctor when a null supplier is provided.

It's easier to remove the constraint in a future commit than to add a constraint, so my preference was to start as strict as possible.

bavardage: will add the checknotnull in withInitialValue

bavardage: done

carterkozak: I think this will need an approval from you, @schlosna, since some of my suggestions have been merged. I'm going to approve for now given the null check is currently set in the direction of least regret, but I'd like to confirm that you're happy with the current approach -- happy to change/revisit if you have a strong opinion.

carterkozak (approved):

PR #1008, opened 2023-03-06
add trace local observers

bavardage:
first follow-up to #1007 

==COMMIT_MSG==
Allow trace locals to be constructed with an observer, which gets called with the value of the trace local (if set) when a trace is complete.
==COMMIT_MSG==

## Possible downsides?
The actual triggering logic here is going to be fiddly, so will do that in a follow-up. There's no great way currently to know when completing a particular span means that a trace is 'done'


tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -138,6 +147,24 @@ public T remove() {
         return (T) traceLocals.remove(this);
     }
 
+    void onTraceComplete(TraceState traceState) {
```
bavardage, line 151: we take a trace state explicitly here because we're e.g. potentially going to be calling this from a detached span

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -149,4 +176,30 @@ public int hashCode() {
...
+
+        @SuppressWarnings("checkstyle:HiddenField")
+        public Builder<T> initialValue(@Nonnull Supplier<T> initialValue) {
+            this.initialValue = Preconditions.checkNotNull(initialValue, "initial value supplier must not be null");
+            return this;
+        }
+
+        @SuppressWarnings("checkstyle:HiddenField")
+        public Builder<T> observer(@Nonnull TraceLocalObserver<T> observer) {
+            this.observer = Preconditions.checkNotNull(observer, "trace local observer must not be null");
```
carterkozak: We generally use the following pattern to avoid hiding fields:
```suggestion
        public Builder<T> observer(@Nonnull TraceLocalObserver<T> value) {
            this.observer = Preconditions.checkNotNull(value, "trace local observer must not be null");
```

tracing/src/main/java/com/palantir/tracing/TraceLocalObserver.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+import javax.annotation.Nullable;
+
+public interface TraceLocalObserver<T> {
+    void onTraceComplete(String traceId, @Nullable String requestId, T value);
```
carterkozak: We could make this a default method with a no-op implementation, retaining the potential for other observer methods in the future in a consistent way.

This method needs javadoc describing what it's measuring.

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: fair enough, i just mildly preferred the argument names being non-generic

tracing/src/main/java/com/palantir/tracing/TraceLocalObserver.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+import javax.annotation.Nullable;
+
+public interface TraceLocalObserver<T> {
```
carterkozak: We could name this “Observer” and enclose it within TraceLocal. Not a strong opinion either way, but it may help with discovery.

bavardage: yeah fair enough - done

tracing/src/main/java/com/palantir/tracing/TraceLocalObserver.java
bavardage: made default

javadoc incoming...

bavardage: 96fd39e1def974666eadb70be278a001bf0ba9e5 words are hard

I think this'll be expanded a bit in follow-up once we've nailed down exactly what it means for a trace to complete

bavardage:
also added a few tests

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
```diff
@@ -29,14 +29,18 @@
  *
  * A trace local is either set (has a value) or unset (has no value).
  *
- * Outside of a trace (i.e. when {@link Tracer#hasTraceId()} is false) then the trace local is always unset.
```
carterkozak: I'm curious why this was removed, I think it provides helpful context to new users.

carterkozak:
I'm happy with the API in this PR, could you PR the plumbing implementation into this PR?

tracing/src/main/java/com/palantir/tracing/TraceLocal.java
bavardage: sorry, not sure what happened - reverting

bavardage:
discussed - we can actually handle the initial usecases without observers, which avoids the whole problem of having to figure out when a trace is 'done' - which can be tricky for a variety of reasons, but e.g. you can have async child spans still ongoing when the 'root' span finishes)

PR #1016, opened 2023-03-14
Log at debug when `completeSpan` methods are used with no active trace

carterkozak:
Such states suggest that a bug has occurred closing more spans than were opened.

==COMMIT_MSG==
Log at debug when `completeSpan` methods are used with no active trace
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -640,6 +640,11 @@ public final class Tracer {
             if (trace.isObservable()) {
                 completeSpanAndNotifyObservers(span, tag, state, trace.getTraceId());
             }
+        } else if (log.isDebugEnabled()) {
+            log.debug(
+                    "Attempted to complete spans when there is no active Trace. This may be the "
+                            + "result of calling completeSpan more times than startSpan",
+                    new SafeRuntimeException("not a real exception"));
         }
     }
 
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -640,6 +640,11 @@ public static <T> void fastCompleteSpan(TagTranslator<? super T> tag, T state) {
             if (trace.isObservable()) {
                 completeSpanAndNotifyObservers(span, tag, state, trace.getTraceId());
             }
+        } else if (log.isDebugEnabled()) {
+            log.debug(
+                    "Attempted to complete spans when there is no active Trace. This may be the "
+                            + "result of calling completeSpan more times than startSpan",
+                    new SafeRuntimeException("not a real exception"));
```
schlosna, line 647: should we also do this on line 656 as well when there is no open span?

```
    private static <T> void completeSpanAndNotifyObservers(
            Optional<OpenSpan> openSpan, TagTranslator<? super T> tag, T state, String traceId) {
        //noinspection OptionalIsPresent - Avoid lambda allocation in hot paths
        if (openSpan.isPresent()) {
            Tracer.notifyObservers(toSpan(openSpan.get(), tag, state, traceId));
        } else if (log.isDebugEnabled()) {
            log.debug(
                    "Attempted to complete spans when there is no active Trace. This may be the "
                            + "result of calling completeSpan more times than startSpan",
                    new SafeRuntimeException("not a real exception"));
        }
    }
```

carterkozak: I don't believe that works because `Optional<OpenSpan>` isn't necessarily present unless the span is sampled:
https://github.com/palantir/tracing-java/blob/eb41f5d4da11a65476db9f90628e54a20a0eb2dc/tracing/src/main/java/com/palantir/tracing/Trace.java#L251-L258

We clear the current trace when the final span is completed, so I don't think that would give us additional coverage.

schlosna: good call

schlosna (approved):

PR #1160, opened 2024-02-12
Use CloseableTracer in DeferredTracer

pkoenig10:
## Changes in this PR
- Make `DeferredTracer.withTrace()` public so `DeferredTracer` can be used in try-with-resources blocks. This supports for more flexible uses (ex. non-final variables) and allows users to avoid allocations.
- Modify `DeferredTracer` to use `CloseableTracer` instead of it's separate `CloseableTrace` interface. This is a better API for users because there is a single `CloseableTracer` type. Ideally this would be an interface instead of a concrete class, but we can't break the ABI.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
@@ -28,7 +28,7 @@ import java.util.Map;
 public class CloseableTracer implements AutoCloseable {
     private static final CloseableTracer INSTANCE = new CloseableTracer();
 
-    private CloseableTracer() {}
+    CloseableTracer() {}
 
     /**
      * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
@@ -81,6 +81,7 @@ public class CloseableTracer implements AutoCloseable {
     }
 
     @Override
+    @SuppressWarnings("DesignForExtension")
     public void close() {
         Tracer.fastCompleteSpan();
     }
--- a/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
@@ -25,8 +25,7 @@ import java.util.Map;
  *
  * <p>Usage: try (CloseableTracer trace = CloseableTracer.start("traceName")) { [...] }
  */
-public class CloseableTracer implements AutoCloseable {
-    private static final CloseableTracer INSTANCE = new CloseableTracer();
+public abstract class CloseableTracer implements AutoCloseable {
 
     CloseableTracer() {}
 
@@ -52,7 +51,7 @@ public class CloseableTracer implements AutoCloseable {
      */
     public static CloseableTracer startSpan(@Safe String operation, SpanType spanType) {
         Tracer.fastStartSpan(operation, spanType);
-        return INSTANCE;
+        return DefaultCloseableTracer.INSTANCE;
     }
 
     /**
@@ -75,15 +74,24 @@ public class CloseableTracer implements AutoCloseable {
             @Safe String operation, TagTranslator<? super T> translator, T data, SpanType spanType) {
         Tracer.fastStartSpan(operation, spanType);
         if (!Tracer.isTraceObservable() || translator.isEmpty(data)) {
-            return INSTANCE;
+            return DefaultCloseableTracer.INSTANCE;
         }
         return new TaggedCloseableTracer<>(translator, data);
     }
 
     @Override
-    @SuppressWarnings("DesignForExtension")
-    public void close() {
-        Tracer.fastCompleteSpan();
+    public abstract void close();
+
+    private static final class DefaultCloseableTracer extends CloseableTracer {
+
+        private static final DefaultCloseableTracer INSTANCE = new DefaultCloseableTracer();
+
+        private DefaultCloseableTracer() {}
+
+        @Override
+        public void close() {
+            Tracer.fastCompleteSpan();
+        }
     }
 
     private static final class TaggedCloseableTracer<T> extends CloseableTracer {
--- a/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
@@ -24,7 +24,6 @@ import com.palantir.tracing.api.SpanType;
 import java.io.Serializable;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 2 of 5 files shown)


tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -81,6 +81,7 @@ public static <T> CloseableTracer startSpan(
     }
 
     @Override
+    @SuppressWarnings("DesignForExtension")
     public void close() {
```
schlosna: Given the constructor is package protected, this is probably fine. May be worth comment that subtypes either need to call super to complete span, or if they completely override `close()` they are responsible for completing the span.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -137,21 +136,27 @@ CloseableTrace withTrace() {
         }
 
         if (isObservable && metadata != null && !metadata.isEmpty()) {
-            return new TaggedCloseableTrace(originalTrace, metadata);
+            return new TaggedCloseableTracer(originalTrace, metadata);
         } else {
-            return originalTrace.map(CLOSEABLE_TRACE_FUNCTION).orElse(DefaultCloseableTrace.INSTANCE);
+            return originalTrace.map(CLOSEABLE_TRACE_FUNCTION).orElse(DefaultCloseableTracer.INSTANCE);
         }
```
schlosna: perf nit: to avoid the `Optional` allocation via `map`:

```suggestion
        if (originalTrace.isPresent()) {
            return CLOSEABLE_TRACE_FUNCTION.apply(originalTrace.get());
        }
        return DefaultCloseableTracer.INSTANCE;
```

schlosna (commented):
Overall looks good, couple nits

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -136,61 +134,30 @@ CloseableTrace withTrace() {
...
+                if (metadata != null) {
+                    Tracer.fastCompleteSpan(metadata);
+                } else {
+                    Tracer.fastCompleteSpan();
+                }
+                if (originalTrace.isPresent()) {
+                    Tracer.setTrace(originalTrace.get());
+                } else if (Tracer.hasTraceId()) {
+                    Tracer.getAndClearTrace();
+                }
```
schlosna: not exactly a new problem, but now that we're exposing `CloseableTracer` to consumers outside the library's control, do we need to make `close()` idempotent?

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
```diff
@@ -136,61 +134,30 @@ CloseableTrace withTrace() {
...
-        INSTANCE;
-
-        @Override
-        public void close() {
-            Tracer.fastCompleteSpan();
-            if (Tracer.hasTraceId()) {
-                Tracer.getAndClearTrace();
+        return new CloseableTracer() {
+            @Override
+            public void close() {
```
schlosna:
```suggestion
            private boolean closed = false;

            @Override
            public void close() {
                if (closed) {
                    return;
                }
                closed = true;

```

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
pkoenig10: What do you mean? We already expose `CloseableTracer` to consumers. Today consumers can write:

```java
try (CloseableTracer span = CloseableTracer.startSpan("operation")) {
    ...
}
```

pkoenig10: I agree we should make it idempotent, but I think that's outside the scope of this PR.

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
pkoenig10: I've pushed an additional refactor to clean this up more.

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
pkoenig10: FWIW I'm basically treating `CloseableTracer` like an interface. Maybe I should just make it an abstract class - I don't think that's an ABI break. There are no external subclasses because the constructor was private (and anyone who is subclassing this is doing something they shouldn't be).

tracing/src/main/java/com/palantir/tracing/DeferredTracer.java
pkoenig10: See https://github.com/palantir/tracing-java/pull/1160#discussion_r1486487220

tracing/src/main/java/com/palantir/tracing/CloseableTracer.java
```diff
@@ -25,10 +25,9 @@
  *
  * <p>Usage: try (CloseableTracer trace = CloseableTracer.start("traceName")) { [...] }
  */
-public class CloseableTracer implements AutoCloseable {
-    private static final CloseableTracer INSTANCE = new CloseableTracer();
+public abstract class CloseableTracer implements AutoCloseable {
```
pkoenig10, line 28: Would it be an ABI break to change this to an interface if there are no external implementations?

pkoenig10: The [JLS section on binary compatibility](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html) doesn't seem to make any guarantees here, but https://stackoverflow.com/q/42294217 seems to suggest that this will work.

If we agree, I'd like to change this to an interface.

schlosna: I think [JLS section 13.4.16](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.16) would be most applicable here.

The case I would be worried about are services using libraries compiled against an older `tracing-java` release where this is not an abstract class or interface, and where `tracing-java` version is newer and is abstract or interface leading to a runtime `LinkageError` (e.g. `AbstractMethodError`, `NoSuchMethodError`, or `IncompatibleClassChangeError`).

@carterkozak for thoughts on if this is safe.

carterkozak: iirc making an abstract class concrete is usually fine, but swapping abstract class to/from an interface breaks ABI. I'm not certain off the top of my head about making a concrete method abstract (`close`). It would almost certainly be safe to provide an empty non-abstract implementation of the method, overridden in all implementations.

pkoenig10: Would you prefer if I did that instead of making the class abstract? I wanted to make sure that implementors were forced to implement `close`. Although it's probably less of a concern given that all implementations must be in this library.

pkoenig10:
Closing because `DetachedSpan` is almost always more appropriate for the use cases being described here.

PR #1248, opened 2024-07-02
Enable creating detached span with specified request id

ayinresh:
## After this PR

==COMMIT_MSG==
Enable creating detached span with specified request id
==COMMIT_MSG==

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
+++ b/tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
@@ -19,10 +19,9 @@ package com.palantir.tracing;
 import com.google.errorprone.annotations.MustBeClosed;
 import com.palantir.logsafe.Safe;
 import com.palantir.tracing.api.SpanType;
-
-import javax.annotation.CheckReturnValue;
 import java.util.Map;
 import java.util.Optional;
+import javax.annotation.CheckReturnValue;
 
 /**
  * Span which is not bound to thread state, and can be completed on any other thread.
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/DetachedSpan.java
```diff
@@ -86,6 +86,23 @@ static DetachedSpan start(
         return Tracer.detachInternal(observability, traceId, forUserAgent, parentSpanId, operation, type);
     }
 
+    /**
+     * Marks the beginning of a span, which you can {@link #complete} on any other thread.
+     *
+     * @see DetachedSpan#start(String)
+     */
+    @CheckReturnValue
+    static DetachedSpan start(
```
pkoenig10, line 95: This method isn't public and isn't used anywhere internally, so this change doesn't really do anything.

pkoenig10 (changes requested):
Can you please update the PR description to describe:
- What you are trying to accomplish
- Why you feel this change is the right way to accomplish that, including any alternatives you considered

pkoenig10:
I don't think we want to allow callers to explicitly set the request IDs. This is an implementation details of the tracing library and isn't used anywhere except the span metadata, for which we already have trace IDs to correlate spans in the same trace.

https://github.com/palantir/tracing-java/blob/6fc8360e49ca0bc74d0231ea2754f277a5f3f037/tracing/src/main/java/com/palantir/tracing/TraceState.java#L72-L78

The existing `DetachedSpan` methods here also leak some internal tracing concerns (ex. `forUserAgent`) and are probably a bit misguided.

Instead of exposing the tracing internals to callers, we should provide higher level methods to achieve the workflows we want to support.

pkoenig10:
Are you getting this confused with our internal `_requestId` logging MDC parameter?

PR #1287, opened 2024-09-19
Provide an API to access the detached trace associated with an exchange

carterkozak:
==COMMIT_MSG==
Provide an API to access the detached trace associated with an exchange
==COMMIT_MSG==

Alternative to #1288 which provides more direct access to the underlying `Detached` tracing exchange attachment. Uses can be more obvious as they interact directly with tracing.

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
@@ -38,6 +38,11 @@ public final class TracingAttachments {
      */
     static final AttachmentKey<DetachedSpan> REQUEST_SPAN = AttachmentKey.create(DetachedSpan.class);
 
+    /**
+     * Gets the {@link Detached} trace state which represents the top-level request being processed. This may
+     * be used to apply thread state to code executing outside traced handlers, exchange completion
+     * listeners, for example.
+     */
     @Nullable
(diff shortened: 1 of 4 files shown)


tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
```diff
@@ -27,5 +31,17 @@ public final class TracingAttachments {
     /** Attachment providing the request identifier. */
     public static final AttachmentKey<String> REQUEST_ID = AttachmentKey.create(String.class);
 
+    /**
+     * Detached span object representing the entire request including asynchronous components.
+     * This is intentionally not public, we expose only the {@link Detached} component which critically does not
```
bjlaub, line 36: > we expose only the {@link Detached} component which critically does not
     * include {@link DetachedSpan#complete()} APIs

I assume this is important so that the caller can only observe the span, not inadvertently call `complete()` which is handled by the exchange completion handler?

carterkozak: Yep, the idea is that when we create this span, we register the exchange completion listener to close it -- we don't want user code to inadvertently close the span _before_ the request has a chance to complete. This object can be used to create new child spans, which themselves may be closed at the users discretion.

bjlaub (approved):
lgtm

PR #1297, opened 2024-09-30
broaden the requestTrace attachment accessor

carterkozak:
==COMMIT_MSG==
broaden the requestTrace attachment accessor
==COMMIT_MSG==

This allows us to pull the attachment value from one of the many types which wrap an HttpServerExchange and delegate to its attachments, for example `WebSocketHttpExchange` (not Attachable) and `ServerSentEventConnection` (Attachable).

### DIFF
--- a/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
+++ b/tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
@@ -54,7 +54,7 @@ public final class TracingAttachments {
      */
     @Nullable
     public static Detached requestTrace(HttpServerExchange exchange) {
-        return exchange.getAttachment(REQUEST_SPAN);
+        return exchange.getAttachment(REQUEST_DETACHED_TRACE);
     }
 
     private TracingAttachments() {}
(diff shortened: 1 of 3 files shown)


tracing-undertow/src/main/java/com/palantir/tracing/undertow/TracingAttachments.java
```diff
@@ -38,14 +38,23 @@ public final class TracingAttachments {
...
     static final AttachmentKey<DetachedSpan> REQUEST_SPAN = AttachmentKey.create(DetachedSpan.class);
 
+    /**
+     * The {@link Detached} trace state which represents the top-level request being processed. This may
+     * be used to apply thread state to code executing outside traced handlers, exchange completion
+     * listeners, for example.
+     */
+    @SuppressWarnings("unchecked")
+    public static final AttachmentKey<Detached> REQUEST_DETACHED_TRACE =
+            (AttachmentKey<Detached>) (AttachmentKey<?>) REQUEST_SPAN;
```
carterkozak, line 48: I don't love exposing this directly since it allows the attachment to be removed, and we don't really expect anyone to do that. The risk is small, and this unblocks applying tracing to some additional libraries without consuming that functionality directly into `tracing-undertow`. In particular `WebSocketHttpExchange` wraps an `HttpServerExchange` and exposes its attachments, but doesn't implement `Attachable` or expose the underlying exchange directly.

mpritham (approved):

PR #1365, opened 2025-02-18
add a Parent-Request-Id header

bjlaub:
## Before this PR
Traces created with a span type of `SERVER_INCOMING` generate a requestId, which is useful for observing operations on a service related to handling a particular inbound request. However, that requestId is local to the service itself, and doesn't allow for easily following RPC call chains in observability data. While the traceId does allow for this, it can sometimes be overly noisy and doesn't allow for easily tracing through RPC calls themselves.

## After this PR
add a Parent-Request-Id header

This provides better observability for RPC call chains by having callers include their own generated requestId (if one exists) on new outbound requests. Receivers can log the `Parent-Request-Id` header value. This should allow for observing the operations that generated a request to an endpoint by following `Parent-Request-Id` pointers backwards, without needing to view observability data for the entire traceId, which may be very large.

==COMMIT_MSG==
add a Parent-Request-Id header
==COMMIT_MSG==

## Possible downsides?
Extra header means more bytes on the wire, but should be small.

### DIFF
--- a/tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
@@ -39,4 +39,14 @@ public interface TraceHttpHeaders {
      */
     @Deprecated
     String ORIGINATING_SPAN_ID = "X-OrigSpanId";
+
+    /**
+     * For traces initiated with span type {@link SpanType#SERVER_INCOMING},
+     * this header will be set to the requestId generated for that trace.
+     *
+     * Inclusion of this header on outbound requests allows receivers to potentially observe the operations
+     * for the direct parent of a request received at an endpoint without the need to observe all operations
+     * in the entire trace.
+     */
+    String PARENT_REQUEST_ID = "Parent-Request-Id";
 }
--- a/tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
+++ b/tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
@@ -16,7 +16,7 @@
 
 package com.palantir.tracing.api;
 
-/** Zipkin-compatible HTTP header names. */
+/** HTTP header names used by tracing-java. */
 public interface TraceHttpHeaders {
     String TRACE_ID = "X-B3-TraceId";
     /**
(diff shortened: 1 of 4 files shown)


tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
```diff
@@ -39,4 +39,14 @@ public interface TraceHttpHeaders {
      */
     @Deprecated
     String ORIGINATING_SPAN_ID = "X-OrigSpanId";
+
```
mpritham, line 42: nit: let's remove the comment `/** Zipkin-compatible ... */`? `Parent-Request-Id` and `For-User-Agent` are not [B3 headers](https://github.com/openzipkin/b3-propagation)

tracing-api/src/main/java/com/palantir/tracing/api/TraceHttpHeaders.java
```diff
@@ -39,4 +39,14 @@ public interface TraceHttpHeaders {
...
+    /**
+     * For traces initiated with span type {@link SpanType#SERVER_INCOMING},
+     * this header will be set to the requestId generated for that trace.
```
carterkozak: I'm having trouble parsing this sentence. Perhaps something like:

> The {@code Parent-Request-Id} header represents the requestId of the caller, if present. This provides a direct causal connection between requests up a call stack. This can be an advantage over traceId which provides a broad set of data without information to understand the call pathing unless spans are sampled.

tracing/src/test/java/com/palantir/tracing/TracersTest.java
```diff
@@ -907,6 +907,28 @@ public void testAddTracingHeaders_populates() {
...
+
+    @Test
+    public void testAddTracingHeaders_addsParentRequestId() {
+        // need to set span type to SERVER_INCOMING to generate a requestId
+        Tracer.initTraceWithSpan(
+                Observability.SAMPLE,
+                "defaultTraceId",
+                Optional.of("forUserAgent"),
+                "rootOperation",
+                SpanType.SERVER_INCOMING);
```
carterkozak, line 922: We can capture the requestId value here, and assert equality below rather than `isNotBlank()`

carterkozak (approved):
thanks!

PR #1406, opened 2025-04-09
Fix parameterized test tracing wrongly accumulating spans

lhhong:
## Before this PR

When running test tracing with parameterized tests, the spans accumulate at each iteration leading to confusing traces.

## After this PR

==COMMIT_MSG==
Fix issue with test tracing snapshot accumulating spans in parameterized testing context. Existing parameterized tests will fail due to extra entries in the expected spans that were wrongly added previously.
==COMMIT_MSG==

### DIFF
--- a/tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
+++ b/tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
@@ -18,7 +18,6 @@ package com.palantir.tracing;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tracing.api.Serialization;
 import com.palantir.tracing.api.Span;
 import java.nio.file.Files;
@@ -33,6 +32,7 @@ import java.util.stream.Collectors;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.platform.commons.support.AnnotationSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -40,11 +40,16 @@ import org.slf4j.LoggerFactory;
 final class TestTracingExtension implements BeforeTestExecutionCallback, AfterTestExecutionCallback {
 
     private static final Logger log = LoggerFactory.getLogger(TestTracingExtension.class);
+    private static final Namespace TRACING_EXT_NAMESPACE = Namespace.create(TestTracingExtension.class);
+    private static final String SUBSCRIBER_STORE_KEY = "subscriber";
 
     @Override
     public void beforeTestExecution(ExtensionContext context) {
         Tracer.setSampler(AlwaysSampler.INSTANCE);
-        Tracer.subscribe(context.getUniqueId(), new TestTracingSubscriber());
+        TestTracingSubscriber subscriber = context.getStore(TRACING_EXT_NAMESPACE)
+                .getOrComputeIfAbsent(
+                        SUBSCRIBER_STORE_KEY, _k -> new TestTracingSubscriber(), TestTracingSubscriber.class);
+        Tracer.subscribe(context.getUniqueId(), subscriber);
 
         // TODO(dfox): sample can be modified by other code, we should be try ensure that the trace is always sampled
         // for the lifetime of the test
@@ -55,9 +60,9 @@ final class TestTracingExtension implements BeforeTestExecutionCallback, AfterTe
     @Override
     public void afterTestExecution(ExtensionContext context) throws Exception {
         String name = testName(context);
-        if (!(Tracer.unsubscribe(context.getUniqueId()) instanceof TestTracingSubscriber subscriber)) {
-            throw new SafeIllegalStateException("Expecting TestTracingSubscriber");
-        }
+        Tracer.unsubscribe(context.getUniqueId());
+        TestTracingSubscriber subscriber =
+                context.getStore(TRACING_EXT_NAMESPACE).remove(SUBSCRIBER_STORE_KEY, TestTracingSubscriber.class);
 
         Path outputPath = getOutputPath(name);
         Path snapshotFile = Paths.get("src/test/resources/tracing").resolve(name + ".log");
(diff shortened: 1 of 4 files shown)


tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -55,7 +55,9 @@ public void beforeTestExecution(ExtensionContext context) {
     @Override
     public void afterTestExecution(ExtensionContext context) throws Exception {
         String name = testName(context);
-        Tracer.unsubscribe(context.getUniqueId());
+        if (!(Tracer.unsubscribe(context.getUniqueId()) instanceof TestTracingSubscriber subscriber)) {
+            throw new SafeIllegalStateException("Expecting TestTracingSubscriber");
+        }
```
pkoenig10: This seems unnecessary

lhhong: Will need to cast it to have the getAllSpans method. I can do a direct cast if that's more preferable?

lhhong: Not obvious from this PR diff but `subscriber` is used below.

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -39,12 +40,11 @@
 final class TestTracingExtension implements BeforeTestExecutionCallback, AfterTestExecutionCallback {
 
     private static final Logger log = LoggerFactory.getLogger(TestTracingExtension.class);
-    private final TestTracingSubscriber subscriber = new TestTracingSubscriber();
```
pkoenig10, line 42: This should be kept as a non-final field that we reset before every test. Then we don't need to cast.

pkoenig10 (changes requested):

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
pkoenig10: Sorry, that's not quite right. This needs to be stored in the `ExtensionContext`.

https://junit.org/junit5/docs/current/user-guide/#extensions-keeping-state

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -32,18 +32,23 @@
...
 
     @Override
     public void beforeTestExecution(ExtensionContext context) {
         Tracer.setSampler(AlwaysSampler.INSTANCE);
+        TestTracingSubscriber subscriber = context.getStore(TRACING_EXT_NAMESPACE)
+                .getOrComputeIfAbsent(
+                        SUBSCRIBER_STORE_KEY, _k -> new TestTracingSubscriber(), TestTracingSubscriber.class);
         Tracer.subscribe(context.getUniqueId(), subscriber);
 
         // TODO(dfox): sample can be modified by other code, we should be try ensure that the trace is always sampled
```
lhhong: @iamdanfox Do you happen to remember what these 2 TODOs are? Seems like this PR resolves them?

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -32,18 +32,23 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.platform.commons.support.AnnotationSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 final class TestTracingExtension implements BeforeTestExecutionCallback, AfterTestExecutionCallback {
 
     private static final Logger log = LoggerFactory.getLogger(TestTracingExtension.class);
-    private final TestTracingSubscriber subscriber = new TestTracingSubscriber();
+    private static final Namespace TRACING_EXT_NAMESPACE = Namespace.create(TestTracingExtension.class);
```
pkoenig10: nit: This is in the `TestTracingExtension` class, so the qualifier is not necessary.

```suggestion

    private static final Namespace NAMESPACE = Namespace.create(TestTracingExtension.class);
```

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -32,18 +32,23 @@
...
+    private static final String SUBSCRIBER_STORE_KEY = "subscriber";
```
pkoenig10:
```suggestion
    private static final String SUBSCRIBER_KEY = "subscriber";
```

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -32,18 +32,23 @@
...
 
     @Override
     public void beforeTestExecution(ExtensionContext context) {
         Tracer.setSampler(AlwaysSampler.INSTANCE);
+        TestTracingSubscriber subscriber = context.getStore(TRACING_EXT_NAMESPACE)
```
pkoenig10: nit: Move this up before the `Tracer.setSampler` call

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
pkoenig10: The only TODO that is resolved by this PR is `TODO(dfox): clear existing tracing??`. The other ones are still relevant.

Although there not really anything we can do about `TODO(dfox): sample can be modified by other code`, so we may just want to remove it.

`TODO(forozco): cleanup stale snapshots` should stay.

tracing-test-utils/src/main/java/com/palantir/tracing/TestTracingExtension.java
```diff
@@ -32,30 +32,33 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.platform.commons.support.AnnotationSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 final class TestTracingExtension implements BeforeTestExecutionCallback, AfterTestExecutionCallback {
 
     private static final Logger log = LoggerFactory.getLogger(TestTracingExtension.class);
-    private final TestTracingSubscriber subscriber = new TestTracingSubscriber();
+    private static final Namespace NAMESPACE = Namespace.create(TestTracingExtension.class);
```
pkoenig10, line 43: nit: whitespace

```suggestion

    private static final Namespace NAMESPACE = Namespace.create(TestTracingExtension.class);
```

pkoenig10 (approved):
Thanks!

PR #1489, opened 2025-08-06
Deprecate TraceMetadata.getParentSpanId

pkoenig10:
This has been unused since https://github.com/palantir/tracing-java/pull/778. There is no reason that RPC callees need to see more than the current span ID.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracer.java
@@ -103,7 +103,7 @@ public final class Tracer {
     }
 
     /**
-     * In the unsampled case, the Trace.Unsampled class doesn't actually store a spanId/parentSpanId stack, so we just
+     * In the unsampled case, the Trace.Unsampled class doesn't actually store a span stack, so we just
      * make one up (just in time). This matches the behaviour of Tracer#startSpan.
      *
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/com/palantir/tracing/Tracer.java
```diff
@@ -114,21 +114,13 @@ public static Optional<TraceMetadata> maybeGetTraceMetadata() {
...
 
-        if (trace.isObservable()) {
-            return trace.top().map(openSpan -> builder.spanId(openSpan.getSpanId())
-                    .parentSpanId(openSpan.getParentSpanId())
-                    .build());
-        } else {
-            return Optional.of(builder.spanId(Tracers.randomId())
-                    .parentSpanId(Optional.empty())
-                    .build());
-        }
```
pkoenig10, line 131: The current behavior is inconsistent:
- If the current trace is sampled with no span, then we return an absent value.
- If the current trace is unsampled (with no span), then we return a present value with a made-up span ID.

We should always propagate trace information if we have it. It's unfortunate that `TraceMetadata.spanId` is a required field - really this should be optional.

In practice, I don't expect this to make much a difference since there should always be at least one span.

schlosna (approved):

PR #1494, opened 2025-08-07
Clean up some Tracer helper methods

pkoenig10:
And clean up some duplicative logic in `Trace` by moving it out to `Tracers`.

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Trace.java
+++ b/tracing/src/main/java/com/palantir/tracing/Trace.java
@@ -16,12 +16,8 @@
 
 package com.palantir.tracing;
 
-import static com.palantir.logsafe.Preconditions.checkArgument;
 import static com.palantir.logsafe.Preconditions.checkNotNull;
-import static com.palantir.logsafe.Preconditions.checkState;
 
-import com.google.common.base.Strings;
-import com.google.errorprone.annotations.CheckReturnValue;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tracing.api.OpenSpan;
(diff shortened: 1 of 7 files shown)


tracing/src/main/java/com/palantir/tracing/Trace.java
```diff
@@ -51,49 +47,15 @@ private Trace(TraceState traceState) {
...
-     * Opens a new span for this thread's call trace, labeled with the provided operation. If the return value is not
-     * used, prefer {@link #fastStartSpan(String, SpanType)}}.
-     */
-    @CheckReturnValue
-    final OpenSpan startSpan(String operation, SpanType type) {
-        OpenSpan span = top();
+    abstract void fastStartSpan(String operation, Optional<String> parentSpanId, SpanType type);
 
-        return startSpan(operation, span != null ? Optional.of(span.getSpanId()) : Optional.empty(), type);
-    }
```
schlosna, line 76: I know these are package private, but have you checked for any sneaky callers? I am pretty sure I can remove some of these from tritium at this point, but https://github.com/palantir/tritium/blob/develop/tritium-tracing/src/main/java/com/palantir/tritium/tracing/ReflectiveTracer.java & https://github.com/palantir/tritium/blob/develop/tritium-tracing/src/main/java/com/palantir/tritium/tracing/RemotingCompatibleTracingInvocationEventHandler.java were some legacy back-compat with ancient remoting code.

schlosna (dismissed):

tracing/src/main/java/com/palantir/tracing/Trace.java
pkoenig10: I don't know of anything like this and my searches don't turn up anything.

schlosna (approved):

PR #1497, opened 2025-08-11
Fix tests

pkoenig10:
Fix a small mistake from https://github.com/palantir/tracing-java/pull/718.

### DIFF
--- a/tracing/src/test/java/com/palantir/tracing/TracerTest.java
+++ b/tracing/src/test/java/com/palantir/tracing/TracerTest.java
@@ -34,6 +34,7 @@ import java.util.Optional;
 import java.util.Set;
 import org.assertj.core.util.Sets;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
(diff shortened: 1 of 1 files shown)


schlosna (approved):

tracing/src/test/java/com/palantir/tracing/TracerTest.java
```diff
@@ -58,7 +59,7 @@ public final class TracerTest {
     @Captor
     private ArgumentCaptor<Span> spanCaptor;
 
-    @AfterEach
+    @BeforeEach
     public void before() {
```
schlosna, line 63: It is unfortunate the checkstyle rule doesn't catch this shape of mismatch. We could add an error-prone check for mismatched `@BeforeEach`/`@AfterEach` & `@BeforeClass`/`@AfterClass` annotations, but not sure how many occurrences it would catch & prevent.

https://github.com/palantir/tracing-java/blob/abc119d93909ea292bb2226908316e21d2cd5d61/.baseline/checkstyle/checkstyle.xml#L352-L353

pkoenig10: FWIW I've never seen this particularly mistake before. And I suspect in the vast majority of cases, this would cause test failures.

This test is somewhat special because there is mutable global state shared between invocations of the tests. So this works as long as the first test that is run isn't relying on the trace state being reset.

PR #1516, opened 2025-09-11
New Tracers.wrapSupplierWithAlternateTraceId() method

ash211:
## Before this PR
https://github.com/palantir/tracing-java/pull/375 added a `wrapWithAlternateTraceId` that takes a Callable. However, Callable's `get()` method has `throws Exception` on it, which forces callers to deal with the possibility of checked exceptions that may not exist.

## After this PR
==COMMIT_MSG==
New Tracers.wrapSupplierWithAlternateTraceId() method
==COMMIT_MSG==

An alternative to `Callable<T>` is `Supplier<T>`, which is nearly the same but without the checked exception on the signature.

Now for callers wrapping code that doesn't declare any checked exceptions don't have to deal with its possibility.

More concretely, callers are required to wrap in a try-catch before:

```
try {
    Tracers.wrapWithAlternateTraceId("someTraceId", "operation", Observability.UNDECIDED, () -> "hi")
        .call();
} catch (Exception e) {
    // not actually possible because () -> "hi" does not throw
    throw new RuntimeException(e);
}
```

and now after with this method, they aren't:

```
Tracers.wrapSupplierWithAlternateTraceId("someTraceId", "operation", Observability.UNDECIDED, () -> "hi")
    .get();
```

## Possible downsides?
- A bit more API surface area.
- can't overload the method name to use `wrapWithAlternateTraceId` like the others, because it creates a compilation ambiguity as called out [here](https://github.com/palantir/tracing-java/pull/375#discussion_r356192460). So have to changer the name to be something else. I chose `wrapSupplierWithAlternateTraceId`

### DIFF
--- a/tracing/src/main/java/com/palantir/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/tracing/Tracers.java
@@ -450,8 +450,8 @@ public final class Tracers {
      * instead of any trace already set on the thread used to execute the callable. Each execution of the callable will
      * use a new {@link Trace tracing state} with the same given traceId. The given {@link String operation} is used to
      * create the initial span.
-     *
-     * @see {@link #wrapSupplierWithAlternateTraceId(String, String, Observability, Supplier)}
+     * <p>
+     * See also {@link #wrapSupplierWithAlternateTraceId(String, String, Observability, Supplier)}.
      */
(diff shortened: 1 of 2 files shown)


ash211:
See internal PR `foundry/object-sentinel/pull/12391` for the planned usage.

schlosna (approved):

